PR #68, opened 2015-01-13
Add loop

briancavalier:
Add generalized loop combinator.  This is more general than scan: scan can be implemented using loop, but not vice versa (it would require an extra map step afterward to extract the value)

This is def open to discussion: Have you encountered use cases for it?  I have on several occasions, but I always forget to write them down for times like this :)

Still needs:

- [x] `most.loop` and `Stream.prototype.loop`
- [x] Unit tests
- [x] API docs

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -45,6 +45,8 @@ most.js API
 	* [until](#until), alias [takeUntil](#until)
 	* [since](#since), alias [skipUntil](#since)
 	* [during](#during)
+1. Looping
+	* [loop](#loop)
 1. Consuming streams
 	* [reduce](#reduce)
 	* [observe](#observe), alias [forEach](#observe)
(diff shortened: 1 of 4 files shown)


briancavalier:
Just added docs for `loop`.  A few open questions:

1. An ascii timeline diagram for `loop` doesn't really capture the essence of loop, imho.  It might be valuable to include one to show that `loop` doesn't actually change the event timeline (even though it does change the *values*).  An actual graphical diagram of `loop` would probably be much better.  Thoughts?
2. I wasn't sure where `loop` belongs in relation to other things.  It is certainly related to `scan` (and `reduce`), but is more general.  I put it in it's own section, but I don't think that's the right thing.  Maybe `loop` and `scan` should be put into the same section?
3. Does the moving average example make sense?

briancavalier:
I think this is ready, except for questions above about docs.

unscriptable:
Looking great, except perhaps the example.  :)

The moving average example (as written) seems very un-stream-ish.  Performing an array.reduce() for every event?  Does it make sense to carry forward the sum and count, instead?

unscriptable:
I guess I'm thinking of the *cumulative moving average*: http://en.wikipedia.org/wiki/Moving_average#Cumulative_moving_average

PR #70, opened 2015-01-15
Use benchmarkjs for perf tests

### LINKED ISSUE #69: Port perf tests to benchmarkjs
The perf test infrastructure is pretty good, but it would be good to use a well-known package, like benchmarkjs.  It may also reduce some maintenance on perf tests.

briancavalier:
Switch to benchmark.js for perf tests.

Close #69


briancavalier:
I think this is ready to go.

unscriptable:
:+1:

briancavalier:
Travis seems messed up.  Looking at the log, the build clearly passed, but Travis thinks it's still running.  Merging.

Issue #71, opened 2015-01-26
Using with React


roman01la:
I'm looking for a way of using Most with React, in particular to handle DOM events via React assignment style. Similar solution is presented with [RxJS](https://github.com/fdecampredon/react-rxjs-todomvc/blob/gh-pages/src/views/header.jsx#L16).

My implementation using Most:
```
function handlerToStream (stream, event) {
  try {
    stream.source.sink.event(performance.now(), event)
  } catch (err) {
    stream.source.sink.error(performance.now(), err);
  }
}

let Button = React.createClass({

  _onClick: handlerToStream.bind(null, (() => {

    let stream = most.create(() => {});
    stream.forEach(event => console.log(event.target));

    return stream;
  })()),
  render() {

    return <button onClick={this._onClick}>click</button>
  }
});
```

Is there a cleaner way to make this work?

briancavalier:
Hey @roman01la, good question.  Here's a version of `Button` that seems like it should do what you intended.  It's untested, but seems close enough for discussion :)  Please let me know if this seems like it's on the right track!

```js
let Button = React.createClass({
	componentWillMount() {
		// emitEvent starts out as a noop, but will be re-assigned below
		// when clickStream is activated
		let emitEvent = () => undefined;
		this._onClick = event => emitEvent(event);
		
		// Stash the stream somewhere.  This isn't necessary, but it's not
		// clear why we'd use a stream if we don't allow someone else
		// to observe it.
		this.clickStream = most.create(add => emitEvent = add);
		this.clickStream.drain(); // Create demand to activate most.create callback
	} 
	
	render() {
		return <button onClick={this._onClick}>click</button>
	}
});
```

This example, and the RxJS one are slightly strange cases.  Reactive streams typically facilitate communication between a producer and one or more consumers.  In this case, though, the producer and consumer are, in fact, the same component.  The RxJS example seems to use that producer-consumer-loopback for every view, and AFAICT, no component consumes a stream produced by another component.

The point of that TodoMVC impl seems to be to try to do as much as possible with RxJS, so I can understand why the author may have chosen to do it that way.  Still, it seems like using reactive streams in that way makes the impl more complicated.

[...]

PR #72, opened 2015-01-27
Add initial Publisher impl for discussion

briancavalier:
For discussion. This is an imperative push adapter that makes for easier integration in situations where `most.create` might be clumsy (see #71), or where events may be emitted *and* observers may be added *in the current call stack*.

### DIFF
--- a/lib/scheduler/PropagateTask.js
+++ b/lib/scheduler/PropagateTask.js
@@ -3,6 +3,7 @@
 /** @author John Hann */
 
 var fatal = require('../fatalError');
+var error = require('../sink/sink').error;
 
 module.exports = PropagateTask;
 
--- a/lib/scheduler/PropagateTask.js
+++ b/lib/scheduler/PropagateTask.js
@@ -43,10 +43,6 @@ PropagateTask.prototype.error = function(t, e) {
 	this.sink.error(t, e);
 };
 
-function error(t, e, sink) {
-	sink.error(t, e);
-}
-
 function emit(t, x, sink) {
 	sink.event(t, x);
 }
--- a/lib/scheduler/PropagateTask.js
+++ b/lib/scheduler/PropagateTask.js
@@ -3,7 +3,6 @@
 /** @author John Hann */
 
 var fatal = require('../fatalError');
-var error = require('../sink/sink').error;
 
 module.exports = PropagateTask;
 
@@ -54,4 +53,8 @@ function emit(t, x, sink) {
 
 function end(t, x, sink) {
 	sink.end(t, x);
+}
+
+function error(t, e, sink) {
+	sink.error(t, e);
 }
--- a/lib/sink/sink.js
+++ b/lib/sink/sink.js
@@ -4,6 +4,7 @@
 
 exports.tryEvent = tryEvent;
 exports.tryEnd = tryEnd;
+exports.error = error;
 
 function tryEvent(t, x, sink) {
 	try {
@@ -19,4 +20,8 @@ function tryEnd(t, x, sink) {
 	} catch(e) {
 		sink.error(t, e);
 	}
+}
+
+function error(t, e, sink) {
+	sink.error(t, e);
 }
--- a/lib/sink/sink.js
+++ b/lib/sink/sink.js
@@ -1,0 +1,22 @@
+/** @license MIT License (c) copyright 2010-2014 original author or authors */
+/** @author Brian Cavalier */
+/** @author John Hann */
+
+exports.tryEvent = tryEvent;
+exports.tryEnd = tryEnd;
+
+function tryEvent(t, x, sink) {
+	try {
+		sink.event(t, x);
+	} catch(e) {
+		sink.error(t, e);
+	}
+}
+
+function tryEnd(t, x, sink) {
+	try {
+		sink.end(t, x);
+	} catch(e) {
+		sink.error(t, e);
+	}
+}
--- a/lib/sink/sink.js
+++ b/lib/sink/sink.js
@@ -2,9 +2,11 @@
 /** @author Brian Cavalier */
 /** @author John Hann */
 
+var fatalError = require('../fatalError');
+
 exports.tryEvent = tryEvent;
 exports.tryEnd = tryEnd;
-exports.error = error;
+exports.tryError = tryError;
 
 function tryEvent(t, x, sink) {
 	try {
@@ -22,6 +24,10 @@ function tryEnd(t, x, sink) {
 	}
 }
 
-function error(t, e, sink) {
-	sink.error(t, e);
+function tryError(t, e, sink) {
+	try {
+		sink.error(t, e);
+	} catch(e) {
+		fatalError(e);
+	}
 }
--- a/lib/source/Publisher.js
+++ b/lib/source/Publisher.js
@@ -4,6 +4,7 @@
 
 var Stream = require('../Stream');
 var sink = require('../sink/sink');
+var copy = require('../base').copy;
 
 var tryEvent = sink.tryEvent;
 var tryEnd   = sink.tryEnd;
@@ -19,7 +20,7 @@ function Publisher() {
 Publisher.prototype.event = function(x) {
 	this._checkState();
 
-	var s = this.sinks.slice();
+	var s = copy(this.sinks);
 
 	for(var i=0, rec; i<s.length; ++i) {
 		rec = s[i];
@@ -31,7 +32,7 @@ Publisher.prototype.end = function(x) {
 	this._checkState();
 	this.active = false;
 
-	var s = this.sinks.slice();
+	var s = copy(this.sinks);
 	this.sinks = void 0;
 
 	for(var i=0, rec; i<s.length; ++i) {
@@ -44,7 +45,7 @@ Publisher.prototype.error = function(e) {
 	this._checkState();
 	this.active = false;
 
-	var s = this.sinks.slice();
+	var s = copy(this.sinks);
 	this.sinks = void 0;
 
 	for(var i=0, rec; i<s.length; ++i) {
--- a/lib/source/Publisher.js
+++ b/lib/source/Publisher.js
@@ -5,51 +5,37 @@
 var Stream = require('../Stream');
 var sink = require('../sink/sink');
 var base = require('../base');
+var PropagateTask = require('../scheduler/PropagateTask');
+var scheduler = require('../scheduler/defaultScheduler');
 
 var tryEvent = sink.tryEvent;
 var tryEnd   = sink.tryEnd;
+var error    = sink.error;
 
 module.exports = Publisher;
 
 function Publisher() {
 	this.sinks = [];
+	this.sink = new PublisherSink(this);
 	this.active = true;
 	this.stream = new Stream(new PublisherSource(this));
 }
 
 Publisher.prototype.event = function(x) {
 	this._checkState();
-
-	var s = this.sinks;
-
-	for(var i=0, rec; i<s.length; ++i) {
-		rec = s[i];
-		tryEvent(rec.scheduler.now(), x, rec.sink);
-	}
+	scheduler.asap(PropagateTask.event(x, this.sink));
 };
 
 Publisher.prototype.end = function(x) {
 	this._checkState();
 	this.active = false;
-
-	var s = this.sinks;
-
-	for(var i=0, rec; i<s.length; ++i) {
-		rec = s[i];
-		tryEnd(rec.scheduler.now(), x, rec.sink);
-	}
+	scheduler.asap(PropagateTask.end(x, this.sink));
 };
 
 Publisher.prototype.error = function(e) {
 	this._checkState();
 	this.active = false;
-
-	var s = this.sinks;
-
-	for(var i=0, rec; i<s.length; ++i) {
-		rec = s[i];
-		rec.sink.error(rec.scheduler.now(), e);
-	}
+	scheduler.asap(PropagateTask.error(e, this.sink));
 };
 
 Publisher.prototype._checkState = function() {
@@ -79,4 +65,27 @@ PublisherDisposable.prototype.dispose = function() {
 
 function remove(x, a) {
 	return base.remove(base.findIndex(x, a), a);
+}
+
+function PublisherSink(publisher) {
+	this.publisher = publisher;
+}
+
+PublisherSink.prototype.event = function(t, x) {
+	propagate(tryEvent, x, this.publisher.sinks);
+};
+
+PublisherSink.prototype.end = function(t, x) {
+	propagate(tryEnd, x, this.publisher.sinks);
+};
+
+PublisherSink.prototype.error = function(t, e) {
+	propagate(error, e, this.publisher.sinks);
+};
+
+function propagate(propagateOne, x, s) {
+	for(var i= 0, rec; i<s.length; ++i) {
+		rec = s[i];
+		propagateOne(rec.scheduler.now(), x, rec.sink);
+	}
 }
--- a/lib/source/Publisher.js
+++ b/lib/source/Publisher.js
@@ -3,6 +3,10 @@
 /** @author John Hann */
 
 var Stream = require('../Stream');
+var sink = require('../sink/sink');
+
+var tryEvent = sink.tryEvent;
+var tryEnd   = sink.tryEnd;
 
 module.exports = Publisher;
 
@@ -13,39 +17,45 @@ function Publisher() {
 }
 
 Publisher.prototype.event = function(x) {
-	if(!this.active) {
-		return;
-	}
+	this._checkState();
+
 	var s = this.sinks.slice();
-	for(var i= 0, rec; i<s.length; ++i) {
+
+	for(var i=0, rec; i<s.length; ++i) {
 		rec = s[i];
-		rec.sink.event(rec.scheduler.now(), x);
+		tryEvent(rec.scheduler.now(), x, rec.sink);
 	}
 };
 
 Publisher.prototype.end = function(x) {
-	if(!this.active) {
-		return;
-	}
+	this._checkState();
 	this.active = false;
 
 	var s = this.sinks.slice();
-	for(var i= 0, rec; i<s.length; ++i) {
+	this.sinks = void 0;
+
+	for(var i=0, rec; i<s.length; ++i) {
 		rec = s[i];
-		rec.sink.end(rec.scheduler.now(), x);
+		tryEnd(rec.scheduler.now(), x, rec.sink);
 	}
 };
 
 Publisher.prototype.error = function(e) {
-	if(!this.active) {
-		return;
-	}
+	this._checkState();
 	this.active = false;
 
 	var s = this.sinks.slice();
-	for(var i= 0, rec; i<s.length; ++i) {
+	this.sinks = void 0;
+
+	for(var i=0, rec; i<s.length; ++i) {
 		rec = s[i];
 		rec.sink.error(rec.scheduler.now(), e);
+	}
+};
+
+Publisher.prototype._checkState = function() {
+	if(!this.active) {
+		throw new Error('stream ended');
 	}
 };
 
@@ -66,4 +76,5 @@ function PublisherDisposable(target, sinks) {
 
 PublisherDisposable.prototype.dispose = function() {
 	this.sinks.splice(this.sinks.indexOf(this.target), 1);
+	this.sinks = void 0;
 };
--- a/lib/source/Publisher.js
+++ b/lib/source/Publisher.js
@@ -10,15 +10,15 @@ var scheduler = require('../scheduler/defaultScheduler');
 
 var tryEvent = sink.tryEvent;
 var tryEnd   = sink.tryEnd;
-var error    = sink.error;
+var tryError    = sink.tryError;
 
 module.exports = Publisher;
 
 function Publisher() {
 	this.sinks = [];
 	this.sink = new PublisherSink(this);
-	this.active = true;
 	this.stream = new Stream(new PublisherSource(this));
+	this.active = true;
 }
 
 Publisher.prototype.event = function(x) {
@@ -80,12 +80,12 @@ PublisherSink.prototype.end = function(t, x) {
 };
 
 PublisherSink.prototype.error = function(t, e) {
-	propagate(error, e, this.publisher.sinks);
+	propagate(tryError, e, this.publisher.sinks);
 };
 
-function propagate(propagateOne, x, s) {
-	for(var i= 0, rec; i<s.length; ++i) {
-		rec = s[i];
-		propagateOne(rec.scheduler.now(), x, rec.sink);
+function propagate(propagateOne, x, consumers) {
+	for(var i= 0, consumer; i<consumers.length; ++i) {
+		consumer = consumers[i];
+		propagateOne(consumer.scheduler.now(), x, consumer.sink);
 	}
 }
--- a/lib/source/Publisher.js
+++ b/lib/source/Publisher.js
@@ -1,0 +1,69 @@
+/** @license MIT License (c) copyright 2010-2014 original author or authors */
+/** @author Brian Cavalier */
+/** @author John Hann */
+
+var Stream = require('../Stream');
+
+module.exports = Publisher;
+
+function Publisher() {
+	this.sinks = [];
+	this.active = true;
+	this.stream = new Stream(new PublisherSource(this.sinks));
+}
+
+Publisher.prototype.event = function(x) {
+	if(!this.active) {
+		return;
+	}
+	var s = this.sinks.slice();
+	for(var i= 0, rec; i<s.length; ++i) {
+		rec = s[i];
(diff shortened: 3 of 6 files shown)


briancavalier:
Should `Publisher.prototype.event/end/error` throw if the publisher has already ended, either normally or due to an error--iow when `this.active` is false?  Right now, it will simply return silently, which is somewhat like calling `resolve` on an already resolved Promise.

It's not clear to me which is better for devs.  It's clearly a mistake, but I'm not sure that throwing is the right thing ... perhaps keep emitting errors on the stream?

roman01la:
> if the publisher has already ended

Just to clarify: How does it can be forced to end? By detaching it from event source, e.g. event listener?

briancavalier:
> Just to clarify: How does it can be forced to end? By detaching it from event source, e.g. event listener?

When using a `Publisher`, the caller is really the source of events, and the intent of a publisher is imperative push.  The only way to signal end is for the caller to call `publisher.end` imperatively.  As such, it would be a mistake, for example, to call `publisher.end` twice, or to call `publisher.event` after `publisher.end`, etc.

I'm wondering what the right thing to do in those cases is.  Seems like the choices are:

1. Return silently.  That's what the code as of f847479 does.
2. Throw synchronously.  Possibly heavy handed, but would force devs to structure their code to avoid this class of mistakes.
3. Emit error event.  This seems wrong.  IMHO, no signal (event, error, or end) should ever propagate through a stream that has already ended.

roman01la:
I'm thinking of `Publisher ` as event handler assigned via, let's say `onclick` property. Thus `onclick = null` is similar to `publisher.end()` and so calling `onclick` will throw an exception. Probably option **2** is a good and intuitive behaviour.

briancavalier:
> Thus onclick = null is similar to publisher.end() and so calling onclick will throw an exception.

Ah, ok, interesting analogy.  Another data point is Node writable streams, which throw if you try to `write` after `end`.  The following logs "OOPS write after end":

```js
var fs = require('fs');
var ws = fs.createWriteStream('message.txt');

ws.end('boop\n');
try {
	ws.write(' fuuu');
} catch(e) {
	console.error('OOPS', e.message);
}
```

roman01la:
So does it make sense to throw on writes after `publisher.end`?

:+1: for this approach

briancavalier:
I'm leaning that way, too @roman01la :)

@unscriptable any thoughts so far?

briancavalier:
Ok, `Publisher` throws now if you try to do `event/end/error` after `end` or after `error`.  I feel good about that since it does seem like a developer mistake.

I'm not really crazy about the name `Publisher`.  Others: Emitter, Producer, Sender.  Another line of thinking: Sink, PushSink, PushStream, WritableStream.  I don't like the last two (*Stream) since it really *isn't* a stream, but rather the publisher and publisher.stream represent the producer and consumer "sides" of the stream of values.

Something else we need to work out:  How should someone create a `Publisher`?  Currently, it's not pretty:

```js
var Publisher = require('most/lib/source/Publisher');
var publisher = new Publisher();
```

That's awful for a couple reasons: 1) the module name is very deep inside `lib`, and 2) you have to use `new`.  Besides the fact that I'm not a fan of `new`, it would also be the first required use of `new` in all of most.js's public API.

Here are a couple ideas which hopefully will spur some discussion:

```js
// move module to top-level, provide create() helper
var Publisher = require('most/Publisher');
var publisher = Publisher.create();

// add API to main module
var most = require('most');
var publisher = most.createPublisher();

// combo
var Publisher = require('most').Publisher
var publisher = Publisher.create();
```

Thoughts?

roman01la:
Well, it's definitely about *pushing* data into Publisher and `publisher.stream` will push data at you. IMHO, common names among mentioned are: Emitter, PushStream and WritableStream. WritableStream is more likely from Node.js world. Emitter looks like something that will produce a bunch of things on its own but without mentioning that you can pass data into it. So my vote is for something about pushing :)

The third possible option of creation publisher is looking good :+1:

briancavalier:
@roman01la Yeah, I agree that "push" is the right connotation.  A few more thoughts: Push, Intake, Inlet, Manifold, Producer, Source (good word, but problematic since most.js already uses the word "source" for a different concept)

> The third possible option of creation publisher is looking good

Thanks.  After thinking about it a bit more, I prefer the first option.  Here's why: For browser applications, it makes it very easy for build tools such as browserify/webpack/etc to exclude the `most/Publisher` module entirely if it isn't used.  IOW, it would result in smaller builds.  The second and third options force build tools to include `Publisher` even if it's not used.

roman01la:
>For browser applications, it makes it very easy for build tools such as browserify/webpack/etc to exclude the most/Publisher module entirely if it isn't used.

@briancavalier A good point, didn't think about it!

unscriptable:
I like the root words "push" and "produce".

unscriptable:
I also agree with you two on the other options: throwing sync and `Publisher.create()` (although perhaps `Producer.create()` or `Push.create()` : ) )

briancavalier:
Rebased on latest master

briancavalier:
I think we're making progress here.  Without picking an actual name for this new thing, it seems like we're all at least in favor of an API like:

```js
var Thing = require('most/Thing');
var thing = Thing.create();
```

Here's one potential variant that may be slightly more ES6 / destructuring friendly:

```js
let Thing = import 'most/Thing';
let thing = Thing.newThing();

// with destructuring:
let { newThing } = import 'most/Thing';
let thing = newThing();
```

That may have an advantage: The name `newThing` makes it obvious what it does, vs. `create` which may leave one wondering "create what?" if it's used at some point farther down in the file.  It's also a few chars shorter than `createThing` ;)

andrewwolters:
Hi @briancavalier, I really like the idea of imperatively pushing events onto a stream. Could this be combined with the notion of some sort of event bus? Essentially, a bus just means: all the streams merged into one, but contrary to a merge operation on streams, a bus doesn't know up front which streams to merge.

I've been experimenting with this kind of code:

```javascript
let p = new Publisher();
let bus = p.stream.join();

// plug a click stream into bus
bus.event(someClicks);

// at another place in the future
bus.event(otherSpecialClicks);
```

One could have an application wide bus for input events. Then, from the – e.g. React – component, you'd plug the local event stream into this global bus. Application logic consumes these inputs from the bus.

briancavalier:
@andrewwolters 

> Could this be combined with the notion of some sort of event bus?

Absolutely.  Dynamic stream network reconfiguration is certainly a powerful concept.

Pretty cool that you were able to make a "bus" simply by using `join`.  What I especially like about it is that it's not dependent on `Publisher` or, in fact, even on imperative push at all.  You could do the same thing by simply mapping an existing stream's events to new streams.  IOW, the network *reconfigures itself* in response to events.

Have you seen the [drag-n-drop example](https://github.com/cujojs/most/blob/master/examples/drag-n-drop/main.js)?  It "plugs" and "unplugs" mouse down, move, and up events using `switch`.  It's basically a mini stateful network.

There are plenty of interesting riffs on the dynamic reconfiguration theme to be had by using `join`, `switch` and `flatMap`, especially when you also think about throwing `since`, `until`, and `during` into the mix to declaratively state when the "plugged" streams should unplug themselves.

(One very minor tweak to your example code (probably a typo): `bus.event(...)` should be `p.event(...)`)

briancavalier:
Here are a couple more crazy thoughts.

#### Call it `Event`

Call it `Event`, and change the method name to `emit` (or some other intuitive verb like `add`/`put`/`send`/etc):

```js
let Event = import 'most/Event';
let time = new Event();

// NOTE: time.error and time.end still exist, too
setInterval(() => time.emit(Date.now()), 1000);

return time.stream;
```

#### Make it a function

This is kind of orthogonal to what we call it, but we could make it a bound function (which could be an issue if you create lots of these things, but I don't forsee that).

```js
let event = import 'most/event';
let time = event();

// NOTE: time() is a function, and time.error and time.end still exist, too
setInterval(() => time(Date.now()), 1000);

return time.stream;
```

Just throwing those out there as more brainstorming.  Any thoughts on either/both??

andrewwolters:
Very nice ideas indeed! I am getting very excited about this monadic style of programming in JS and shall certainly give my thoughts.

But first, I have a puzzling behavior with Publisher. For some reason, the joined stream from a Publisher is not merge-able with another stream. Example code is in the whereabouts of:

```javascript
let p = new Publisher();
most.merge(someOtherStream, p.stream.join()).
   tap(x => console.log(x)).
   drain();

p.event(most.of(1));
p.event(most.from([2, 3, 4]));
```

Could you shed some light on this matter? I'm trying to get a grasp of `most`'s internals. Is `p.stream` already closed? Is no one subscribing? Something to do with active vs passive streams? Or task scheduling?

briancavalier:
@andrewwolters 

> I have a puzzling behavior with Publisher

Good catch, really glad you pointed this out.  There is def still some thinking and work to do on `Publisher`.  Imperative push exposes a level of "how" and "when" that wasn't currently exposed before.  In declarative sources, the internal machinery can manage more closely when everything gets "hooked up" so that events don't get lost.

In this case, getting a direct handle to a `Publisher` in the current tick means you can immediately call `p.event()`.  Even though it *seems* like someone is listening to that event, due to the earlier call to `drain`, that's not the case in the code as it stands right now.  The reasons why are subtle atm, but it's something that'll need to be worked out before `Publisher` can go live.

To verify, you can try:

```js
var p = new Publisher();

most.merge(most.periodic(1000, 'hi'), p.stream)
	.observe(console.log);

process.nextTick(function() {
	p.event(1);
});
```

With `process.nextTick` you'll see the `1`.  Without it, you won't.  Isn't asynchrony fun?  Basically, without the `process.nextTick` the `1` is pushed into the stream "too early", and no one is listening.  With computed (unfold, iterate, etc.) and declarative streams (from, fromEvent, etc.), that problem doesn't exist, since the internal machinery manages it all and can rule out the race condition.  For example, you'll also see the `1` here:

```js
most.merge(most.periodic(1000, 'hi'), most.of(1))
	.observe(console.log);
```

So, yep, still some work to do here :)

briancavalier:
Rebased on latest master.

briancavalier:
@andrewwolters BTW, there is a safe imperative push adapter already, if you need it: [`most.create`](https://github.com/cujojs/most/blob/master/docs/api.md#mostcreate).  It doesn't externalize as much control or footgun ammo as `Publisher` does, and for certain imperative cases it's harder to use.

briancavalier:
I just pushed another experiment.  This one enqueues all events, releasing all queued events on the next microtick.  This will help with the most common (imho) race condition, which is current-tick push vs. next-microtick consumer attachment.

An interesting variation of the scenario is:

```js
p.event(most.of(1));
p.event(most.from([2, 3, 4]));

most.merge(someOtherStream, p.stream.join())
	.tap(console.log)
	.drain();
```

No one is consuming when the two events are pushed, so they are never logged.  To me, that seems perfectly reasonable (think of how other event systems, like the DOM, work, for example).  I think the only real alternatives in that case are:

1. Wait "longer" (arbitrary) before propagating initial events, or
2. Queue events, *potentially forever*, waiting for a consumer.  This seems far too dangerous (again, imagine if DOM events worked this way!)

Neither of those is appealing.  I'm definitely open to ideas/thoughts/suggestions :)

roman01la:
@briancavalier Hey! What's the state of Publisher? Is going to be merged into master? :)

briancavalier:
Hey @roman01la, I've been distracted by other things for a while :)  Now that 0.13.0 has landed, I'll make time for Publisher.  I think the latest experiment is a really good candidate for merging to master, so my plan is to review and clean it up for final review within the next week.

briancavalier:
@roman01la Ok, I'm finally back to looking at Publisher again for real.  I did some perf testing on it, and just pushed a pretty big perf improvement.  Next I need to add some docs and tests.

nissoh:
Hey @briancavalier. i fiddled abit around the next major version of RxJS.
https://github.com/ReactiveX/RxJS
```
    import {Observable} from '../../lib/RxJS/observable/observable'

    var observable = new Observable((generator) => {
      console.log(1);
      generator.next(2);
    });

    observable.observer({ // es6 generator object
      next (value) {
        console.log(value);
      }
    });

    console.log(3)

    // 1 2 3
```

The new version seems to be somewhat compatible with this ES7 proposal of 
https://github.com/jhusain/asyncgenerator

Also, it's important to note react's proposed usage of observables. alleged to be released in the next release 0.14
https://github.com/facebook/react/issues/3398#issuecomment-90679874

I hope this helps (:

briancavalier:
Thanks, @nissoh!  I've been following @jhusain's proposal for a while.  It seems to be moving along, although I do have some concerns about it (I've talked to him about them).  Thanks also for the pointer to the react discussion.

It's still fairly far off, so it's hard to say where ES7 Observable will end up, ultimately.  It still may be worthwhile to try to add support of for the in-progress proposal to find issues, etc.  It should be easy.  I think I'll whip up strawman PR for discussion.

briancavalier:
Hmmm, if react can support the current Observable proposal in the short term, then Publisher may not be needed.  Any thoughts on that @roman01la?

briancavalier:
@nissoh I just noticed that in your example 3 is logged *after* 1 and 2.  This is one of the issues I've expressed to @jhusain.  If the producer is synchronous, 3 will be logged last, but if it's asynchronous, 3 will be logged *first*.  IOW, events may be emitted *before the Observable constructor ever returns*.  

That "maybe first, maybe last" behavior, aka zalgo, seems far too hazardous to me.  I feel quite strongly that no event should ever be emitted before the Observable constructor returns.  That would enforce that the order is 3 1 2 regardless of whether the producer is sync or async.

It's still possible to get high performance in that case, by simply waiting for the stack to clear and then allowing fully synchronous propagation (that's what most.js does now).

briancavalier:
Well, that was easy :) See the [add-es7-observable branch](https://github.com/cujojs/most/tree/add-es7-observable).  It doesn't support everything in https://github.com/jhusain/asyncgenerator but it's pretty close.

Here's a modified version of @nissoh's example that works with the branch:

```js
var Observable = require('most').Observable;

var observable = new Observable(function(generator) {
	console.log(1);
	generator.next(2);
});

observable['@@observer']({
	next: function (value) {
		console.log(value);
	}
});

console.log(3);

// Logs 3 1 2
```

briancavalier:
See #118 for more discussion on ES7 Observable compatibility.

briancavalier:
Rebased on latest master.

briancavalier:
I can think of 3 paths we can take:

1. Don't merge this.  Let the es7 observable direction play out, or
2. Change Publisher API to match ES generator API (i.e. `next`, `return`, and `throw`) and merge, or
3. Merge as is

I'm definitely seeking feedback.

I don't want to create a new API that becomes obsolete if react is already planning to support an es7 observable style API.  However, if that is going to take a while to be practically usable, it may still make sense to release Publisher.

Thoughts?

jhusain:
Hey Brian,

I would let things play out. In a couple of meetings I expect that we will get a good idea of whether this type will be successful. We will also see if any changes are hammered into it during the committee process. For now the fact that angular and react are using something similar to the proposal is sufficient to demonstrate demand. 

JH

briancavalier:
Thanks @jhusain, really appreciate the input.  I'm currently leaning that way as well.

nissoh:
I think releasing Publisher & adopting observable will cause some bad habits.

+1 on forcing asynchrony.. meaning to dispose all of this (:

romeovs:
What's the status on this? I kind of need a Publisher for this project I'm building (w/o React, believe it or not), and I'm trying to stay away from implementing one myself. I'd like to use `most` because I like it a lot but I cannot justify it if it does not have a means of pushing events to a stream easily.

briancavalier:
Hey @romeovs, have you looked at `most.create`?  It provides a safe way of imperatively pushing events.  The reason it is "safe" is that it helps avoids missed events by only running the producer function once the first observer has been added.  IOW, it doesn't allow you to accidentally push an event before someone is listening.  Publisher allows that, and it can be a confusing situation to debug.

That said, I'm still not against adding something like Publisher if there are good use cases for it.  Could you describe your use case in more detail?  Thanks!

romeovs:
I do not see how to push things to the stream using `most.create` from anywhere else than from the body of the passed function. I can see why this is useful and safe in cases where you do not need to have an external pushing mechanism.

My usecase is the following: I'm trying to see if I can replace `RxJS` with `most` in this new stream-bases UI framework [Cycle.js](https://github.com/staltz/cycle).  Because of the way `cycle.js` works, I need to have a circular stream:

```
+--> interactions --> intentions --> vtree ---+
|                                             |
+---------------------------------------------+
```

The way it is solved in the framework now is the following (conceptually):
```js
var interactions1 = emptyStream();
var intentions    = interactions.map(interactionToIntention);
var vtree         = intentions.map(intentionToVtree);
var interactions2 = vtree.map(getInteractions);

// tie the knot
interactions2.forEach(function (ev) {
  interactions1.push(ev);
});
```

This is a pretty sane way to solve it, but it wouldn't work because with `most` all of this code would basically live in the body of the `most.create` callback.  This would be ok, but we need a mechanism of returning the `vtree` stream too.

I might be able to rewrite the framework like that, but this just seems not so easy w/o an imperative Publisher. (I'm still quite new to reactive programming).

Another uselecase is that we want to provide users of the framework a way to create their own events
and pushing them into the cycle.

romeovs:
I tried this to create a cycle:
```js
import most from 'most';

const log = console.log.bind(console);

var cycle = most.create(function (add) {
  cycle
    .startWith(1)
    .observe(function (el) {
      log('adding', el);
      add(el + 1);
    });
});

cycle
  .observe(el => log('saw', el))
  ;
```
but it didn't work, sadly enough. I'd expected it to go on forever (every element is pushed on again in the callback).

pozadi:
@romeovs you can extract `add` functions from the callback pretty easy:

```js
let add;
let stream = most.create((_add) => {
  add = _add;
  return () => {
    add = undefined;
  };
});

// Now you can add values to the stream outside of the callback
if (add !== undefined) {
  add(1);
}
```

It's a bit awkward, but you should't need this often.

romeovs:
the only way I can get that to work in a cycle is using `setImmediate`:
```js
import most from 'most';

var cycle = function (computer) {
  let add;
  const stream = most.create(function (_add) {
    add = _add;
  });

  const computed = computer(stream)

  computed
    .observe(function (el) {
      setImmediate(function() {
        add(el)
      });
    });

  return computed;
};


var computer = function (stream) {
  const res = stream
    .startWith(0)
    .filter(el => el < 10)
    .map(el => el + 1)
    ;

  res.observe(s => s);

  return res;
};

var s = cycle(computer);

s.observe(el => console.log('saw', el));
```

pozadi:
Well, you are creating an infinite loop without `setImmediate`, what do you expect :)
Cycle.js also don't simply connect input and output of the stream. It has a layer in between that converts stream of virtual-dom trees to stream of user events. In this layer infinite loop breaks in Cycle.js, as events in __stream of virtual-dom trees__ don't cause events in __stream of user events__ directly.

romeovs:
I know this would create an infinite loop. But that's kind of what I wanted 😊. I wanted to see if I could do that.

I know cycle works differently, but I was planning on starting from the infinite loop case.

pozadi:
But infinite loop will just freeze the browser.

romeovs:
Yes I know, but my final code won't be an infinite loop. It will be a cycle however. I just wanted to see if I could get a cycle and an infinite loop is the most basic cycle. I'm adding the event filtering layer now.

briancavalier:
@romeovs Cool, thanks for explaining.  If you'll be dealing with DOM events at some point in the cycle, that will naturally remove the synchronous infinite loop to create an asynchronous cycle.

It seems like you're on a good path now using `most.create`.  Please keep us posted as to whether you think `most.create` will work for you after you get further along.

romeovs:
I just bumped into a new use case where I needed this again.
I looked at your code example in #71 to further understand the problem.
You examples basically always wind down to this:
```js
emit = function ( ) { }; // initially drop events
var stream = most.create(function (add) {
  emit = add;
});
stream.drain();
```
I realise this possibly drops early events. Are there any other problems using this approach?

romeovs:
also. I can't get the examples to work when `stream.drain()` is used.

briancavalier:
> I realise this possibly drops early events. Are there any other problems using this approach?

None that I know of :D  This approach is no more problematic than using Publisher.  Either way, there is the possibility of missing events.

On a related note, the reason `most.create` works the way it does is so it can manage the start/stop of the producer function in response to *demand*.  In a purely imperative approach (like `Publisher`) it's actually quite a bit easier to hit the problem of missing early events because you'll never know when there's demand and where there isn't.  When used as intended, you literally cannot miss early events with `most.create`, because it won't let you push any until at least one consumer is observing the stream.

> also. I can't get the examples to work when stream.drain() is used.

Can you post one of the examples that shows the issue?

One thing to keep in mind: `most.create` always invokes the producer function asynchronously.  So, in your snippet above, `emit` will not be set to `add` until at least one microturn after you call `stream.drain`.  For example, the follow will only log `2` and will not log `1`

```js
let push = () => {};
let s = most.create(add => push = add);

s.observe(x => console.log(x));

push(1);

setTimeout(() => push(2), 0);
```

Hope that helps.

romeovs:
> When used as intended, you literally cannot miss early events with most.create, because it won't let you push any until at least one consumer is observing the stream.

That actually makes sense to me now! I just put up the relevant code [here](https://github.com/romeovs/borges/blob/master/src/server/events.js). I use a buffer to cache the missed events and replay them when someone is listening.

>> also. I can't get the examples to work when stream.drain() is used.

> Can you post one of the examples that shows the issue?

If you add `events.drain()` at line of the above code for example, this won't work anymore :(

briancavalier:
> That actually makes sense to me now!

Cool :)

> I use a buffer to cache the missed events and replay them when someone is listening

Yeah, that should work.  A thoughts ... sorry if these aren't relevant to your use case, but they just popped into my head while looking at the code:

- You may want to use a rolling queue, dropping older events and retaining newer ones instead of dropping the entire queue once it is full.
- You may want to consider the time of events.  For example, if events are really old ("old" is subjective and probably depends on your use case), you may just want to drop them.  For example, consider mousemove events.  It may not make sense to replay mousemove events that are several seconds old.

> If you add events.drain() at line of the above code for example, this won't work anymore :(

I'm still not sure I understand.  Here is my previous example, with `s.drain()` added.  It still works as I expect it to: it logs `2`.

```js
let push = () => {};
let s = most.create(add => push = add);

s.observe(x => console.log(x));

push(1);

setTimeout(() => push(2), 0);
s.drain();
```

Could you please post a runnable, modified version that shows the problem you're describing?  Thanks!

romeovs:
> - You may want to use a rolling queue, dropping older events and retaining newer ones instead of dropping the entire queue once it is full.
I thought of that, but the events are not *that* important, so I'm okay with dropping them.

> Could you please post a runnable, modified version that shows the problem you're describing? Thanks!
Correct me if I'm wrong but in the example code you gave, the `s.drain()` is superfluous because
you called `s.observe`, so there is already demand to get the stream going. The example also works
if you don't add the `s.drain()` line.

A more useful example would be to use `.tap` to show what's happening on the stream
without creating demand.  Now the `s.drain()` line should be the one who creates demand,
but this fails to work:
```js
let push = () => {};
let s = most.create(add => push = add);

s.tap(x => console.log(x));

push(1);

setTimeout(() => push(2), 0);
s.drain();
```
This example outputs nothing.

## More about the use case:
The problem is that I'm just creating the stream and exposing it to the users of a library I'm writing, so
I have no control over if and when they might decide to start observing the stream. Therefore, to avoid dropping events, I need to create demand myself. I  can do this using `s.observe`.

Now another problem remains: I do not want to defer all events that put something on the stream
until the stream is "initialised" (here I mean `push` is available). I solved this by implementing the 
rudimentary buffer you [already looked at](https://github.com/romeovs/borges/blob/master/src/server/events.js), but that feels hacky.

After writing the above, I realised I probably do need to defer the code that creates stream events, 
since there is no way to  create the stream in completely synchronous way (without using `setTimeout`).

I guess that leaves just one problem: why does the `s.drain` not work in my examples?

briancavalier:
@romeovs Thanks for posting your example.  I see the issue.  In your example, you are draining `s` instead of draining the stream returned by tapping `s`.  Streams in most.js are immutable.  So, `s.tap` does not mutate `s` by adding a tap to it.  Rather, it returns a new stream which is a tapped view of `s`.

So, in your example, `s` *is* being drained, but you can't see it.  The stream returned by `s.tap` remains inert in the example.

Here's a modified version of the example:

```js
let push = () => {};
let s = most.create(add => push = add);

var tapped = s.tap(x => console.log(x));

push(1);

setTimeout(() => push(2), 0);
tapped.drain();
```

Hope that helps!

romeovs:
@briancavalier wow. ok. How silly of me!

briancavalier:
@romeovs no worries, dude!

briancavalier:
@TylorS's [most-subject](https://github.com/tylors/most-subject) can be used as an imperative publisher.

PR #76, opened 2015-02-09
Add value to most.periodic()

### LINKED ISSUE #74: Deprecate repeat(x)
It seems at best useless, and at worst dangerous :)

It's basically a continuous constant signal, but implemented as discrete events, which is problematic: infinitely many occurrences of x at infinitely small time intervals O.o ... it will happily blast out xs as fast as it possibly can.

Some better options might be:

1. Change `repeat` to take a time interval: `repeat(period, value)`
2. Change `periodic` to take a value (or an optional value, :/ )
3. Invent a new operation that takes a time interval and value

briancavalier:
This deprecates `most.repeat` in favor of adding a value to `most.periodic`, eg `most.periodic(period, x)`.

Close #74

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -176,6 +176,8 @@ stream.take(100)
 
 ### most.repeat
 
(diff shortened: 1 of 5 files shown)


unscriptable:
:+1:

PR #77, opened 2015-02-10
Add most.generate to compute a stream from an async generator

briancavalier:
This computes a (potentially infinite) stream, similarly to `most.unfold` or `most.iterate`, from an *async* generator--that is, a generator that yields promises.  This allows the generator to control event times.

For example, compute an infinite stream of random values at random intervals:

```js
function* randoms() {
	for(;;) yield new Promise(resolve => setTimeout(() => resolve(Math.random()), 1000 * Math.random()));
}

var stream = most.generate(randoms);
```

Still needs:

- [x] Unit tests
- [x] API docs

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -14,6 +14,7 @@ most.js API
 	* [most.never](#mostnever)
 	* [most.iterate](#mostiterate)
 	* [most.unfold](#mostunfold)
+	* [most.generate](#mostgenerate)
 	* [most.fromEvent](#mostfromevent)
 	* [most.fromEventWhere](#mostfromeventwhere)
 	* [most.create](#mostcreate)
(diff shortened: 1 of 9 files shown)


briancavalier:
Another example, to show the error handling capabilities.  This produces a stream of 2 events, whose values are `1` and `2`.

```js
function* g(x) {
	yield x;
	try {
		yield Promise.reject(new Error('oops'));
	} catch(e) {
		yield x+1;
	}
}

most.generate(g, 1)
	.observe(console.log);
```

briancavalier:
My example in the docs is kind of unsatisfying :(  Anyone have any suggestions/ideas?

briancavalier:
I simplified the doc example.  I think it's fine for now, but something better would be, well, better ;)

PR #80, opened 2015-02-15
Add sample(f, sampler, streams)

### LINKED ISSUE #78: Rethink sampling
Right now we have `stream.sampleWith(sampler)`, which works nicely.  It's fairly narrow in that it allows sampling one stream, and passing through the value directly.

One question this raises is how to sample 2 streams or N streams.  Say you wanted to add the values of two inputs, and display the result *only when the user clicks a button*.  You could `combine` the inputs, and then sample the resulting stream:

```js
most.combine(add, s1, s2).sampleWith(buttonClick);
```

To cut down the noise, you might do:

```js
most.combine(add, s1.distinct(), s2.distinct()).sampleWith(buttonClick);
```

But that's still doing extra, unnecessary `add`s: each time `s1` or `s2` changes, `add` will be called, even while the user isn't clicking the button.  If `add` is trivial, then you probably don't care.  On the other hand, it *is* doing extra work and throwing it away, which could be especially bad if `add` is nontrivial.

Another option might be to flip the whole thing around in a "sample that's like combine" combinator.  Like:

```js
buttonClick.sample(add, s1, s2);
```

That allows sampling multiple inputs, and would only call `add` when the button is actually clicked, no matter how often `s1` and `s2` change.

Interestingly, it seems like both `combine` and `sampleWith` could be reimplemented using `sample`, although it's not clear the perf impact this may have on `combine`.

```js
function combine(f, ...streams) {
    return merge(streams).sample(f, streams);
}

function sampleWith(sampler, stream) {
    return sampler.sample(identity, stream);
}
```

briancavalier:
Docs and unit tests included.

Close #78


briancavalier:
It's not clear to me why github lists travis as having 1 failing check.  If you click through, it's all green :/

PR #84, opened 2015-02-18
Make applicative more sensible

### LINKED ISSUE #73: Rethink Applicative
Right now `stream.ap` is defined as a cross-product.  This is a classic definition for finite lists, but is kind of useless for infinite streams: given `fs.ap(xs)` where `fs` is an infinite stream of functions, and `xs` is an infinite stream of values, you simply end up with the same result as `xs.map(f0)` where `f0` is the first function in `fs`.  So, the cross-product is basically impossible to observe:

`fs.ap(xs) -> [f0(x0), f0(x1), f0(x2), ...]`

A potentially more useful definition might be a zip-like applicative, ie pair-wise applying the `fs` to the `xs`.  That is:

`fs.ap(xs) -> [f0(x0), f1(x1), f2(x2), ...]`

briancavalier:
This is a much more sensible behavior for applicative.  The previous cross product (which is standard for a finite, synchronous list) had the very serious issue of observing the `xs` stream many times.  The new approach applies the latest `f` to the latest `x`, when either a new `f` or new `x` arrives.  Or, put another way, it applies a *time-varying function* to a *time-varying value*, which is a fairly intuitive definition, imho.

The standard ap laws still pass, which further reinforces that this is a valid approach.

Note that this changes the behavior of combine such that it only ends once all input streams end.  That's potentially a breaking change, so will go out in 0.12.0.

Close #73

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -635,15 +635,15 @@ most.from([1, 2])
 ####`streamOfFunctions.ap(stream) -> Stream`
 ####`most.ap(streamOfFunctions, stream) -> Stream`
 
-Apply all the functions in `streamOfFunctions` to all the values in `stream`.
+Apply the latest function in `streamOfFunctions` to the latest value in `stream`.
 
 ```
-streamOfFunctions:            f-g-h|
-stream:                       x-y-z|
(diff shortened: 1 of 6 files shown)


unscriptable:
although I haven't reviewed the code, I agree with the principles: "applies a time-varying function to a time-varying value"

Issue #86, opened 2015-03-03
Consider deprecating most.fromEventWhere


briancavalier:
The main motivation for `most.fromEventWhere` was that prior to 0.10, event propagation was async, and so `most.fromEvent(...).tap(e => e.preventDefault())` did not work.

Since 0.10, event propagation is synchronous whenever possible.  In particular, in constructs like `most.fromEvent(...).filter(f)`, or `most.fromEvent(...).tap(f)`, `f` will always be called in the call stack in which events originate.  That means, for example, `most.fromEvent(...).tap(e => e.preventDefault())` will work as expected.

Should we deprecate `most.fromEventWhere`?

unscriptable:
In what situations might a lambda (predicate, etc) be called async?  Is it only when the dev introduces a promise into the stream?

briancavalier:
The only situation is when you explicitly use a combinator that is naturally async, like `delay`, `await`, etc.  Other combinators are always sync.  Some are less obvious than others, like `flatMap(f)`, which is async because `f` must return a stream, and streams always start async.

However, one key in this specific case is that using `filter` or `tap` immediately "to the right" of `fromEvent` will *always* be in the same call stack.  Those are the two you'd most likely use to do a selector match or `preventDefault`.  Doing them in succession would also always be sync: `domEvents.filter(selectorMatch).tap(preventDefault)` ... i.e. only preventing default on matching events.

briancavalier:
Deprecated in #91.

PR #91, opened 2015-03-05
Deprecate fromEventWhere

briancavalier:
I haven't thought of a reason *not* to deprecate it.  We could release this deprecation in 0.13 and remove for either 0.14, or 1.0.  If we think of a reason to keep it between now and then, we can reinstate it.

Still needs:

- [x] Update docs

See #86

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -325,13 +325,44 @@ var clicks = most.fromEvent('click', document.querySelector('.the-button'));
 // We can do some event delegation by applying a filter to the stream
 // in conjunction with e.target.matches this will allow only events with 
 // .the-button class to be processed
-var clicks = most.fromEvent('click', document.querySelector('.container'));
-clicks.filter(function(e){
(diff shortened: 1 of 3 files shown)


unscriptable:
The docs need some attention too.  Seems like the two sections, fromEvent and fromEventWhere could be combined easily.  I might be able to tackle that tonight.

briancavalier:
Good call.  That'd be awesome if you combine them!

briancavalier:
Docs updated, and added a unit test to ensure synchronous propagation, which was the premise of deprecating `fromEventWhere` in the first place :)

unscriptable:
Looks like a perf test snick in there.  The rest is :+1:! :)

briancavalier:
Oh man, good catch! Thanks!

briancavalier:
Removed the accidentally added perf test and squashed commits.

Issue #85, opened 2015-02-18
Explore external fusion/rewrite rules, compiler


briancavalier:
Many pairs, triples, etc. patterns of combinators can be simplified algebraically.  For example, `s.map(f).map(g)` is equivalent to `s.map(g•f)`, where `•` is function composition.  Thus a sequence of N `map` operations can be reduced to a single `map`.  Similarly, a sequence of N `filter` operations can be reduced to a single `filter` using predicate composition (ie `p(x) && q(x)`).

More sophisticated simplifications are possible as well.  For example, `s.filter(p).map(f)` can be simplified to a "filtermap" operation, which itself then composes on the left with additional `filter` ops and on the right with additional `map` ops.  IOW `filtermap.map` can be reduced again to a single `filtermap`.

Rather than have each combinator know how to combine itself with any other combinator, I think these rules should be applied from the outside somehow.  For example, some sort of "compiler" which can analyze and "rewrite" combinator sequences.  It could be built from an extensible set of matching rules, where each rule is able to recognize a specific combinator sequence, such as `.map.map` and substitute a simplified/optimized replacement.

briancavalier:
I've done some initial experiments, and the performance benefits can be significant in some cases.  For example, adjacent filter predicate composition seems to make a huge difference.  In other cases, though, the gains are less significant, such as adjacent map composition.  Still others seem to actually make things worse (maybe my test implementations were crummy), such as turning `.map.filter` into a "mapfilter".

briancavalier:
[Here are some initial results](https://gist.github.com/briancavalier/9393d0c9e9e4055629c8) run on io.js 1.4.2.  This test used fusion rules for map.map, filter.filter, and filter.map, and can fuse any number of adjacent filter/map.  For example, the last run in the gist fuses filter.filter.map.map, i.e. 4 steps, into a single step.

unscriptable:
Nice! I did not suspect this much performance gain. It seems like filter fusion is a especially substantial benefit. 

So… what are the downsides?  How's the performance of the fusion-compile step? Is it negligible?  Is there a way to disable the fusion step for CSP environments?  Do we care about CSP environments? :)

unscriptable:
After re-reading the OP, I may have misunderstood. Are you fusing the source codes of the filter and nao functions? Or are you just using some
form of function composition?

briancavalier:
> Are you fusing the source codes of the filter and map functions? Or are you just using some form of function composition?

[...]

PR #92, opened 2015-03-06
Implement filter + map fusion

briancavalier:
This implements the filter.filter, map.map, and filter.map fusion from #85.  I moved the Map and Filter sources to lib/fusion, since Map needs to import Filter now.  I also added FilterMap which fuses a filter and a map.  It can be further fused with subsequent map steps on the right.

I'm still not crazy about the `instanceof` checks, but they're fairly obvious, and I can't see a better way atm--definitely open to suggestions (`source.constructor === Map` ???).  The impact of the `instanceof` checks shouldn't even be noticeable since they happen at stream creation time.  The benefit of the fusion at propagation time far outweighs it.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,7 +1,7 @@
 {
 	"browser": true,
 	"node": true,
-	"predef": [ "-Promise" ],
+	"predef": [ "-Promise", "-Map" ],
 	"boss": true,
 	"curly": true,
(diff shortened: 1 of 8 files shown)


unscriptable:
I can't think of anything better, atm.  Looks good to me. :)

unscriptable:
woot :tada:

Issue #93, opened 2015-03-07
Are there any plans for `fromNodeStream` or a `pipe`?


nackjicholson:
First of all, I've tinkered around with RxJS, as well as Highland, and I prefer most because of the way it nails the interaction with promises. Also, your docs with the stream diagrams are just miles above the other two.

Second, I'm using most in nodejs, and I'm wondering if you have any thoughts about whether node streams have a future in most. I've been able to use `most.create` to get stream data through a most stream, but there really needs to be a way to pipe a most through another node stream. As well as ways to handle back pressure. Otherwise it's not really very useful, reading from a large node stream is like a firehose, and it will fill up the internal buffer and grind node to a halt after a while.

Have you had any thoughts about adapting most to node streams? Is there any priority on additions like this? I am happy mixing and matching with highland, but like I said -- I much prefer most, and it would be great to be able to do things like this with it. Maybe if I have some time I'll try to learn the concepts you have around "sources" and "sinks" to see if I can fork this repo and add something. Any guidance on that front would be greatly appreciated.

Cheers!

nackjicholson:
Here is a link to an open issue RxJS has about the same issue in case you're curious. They have node adapters, but they don't solve the large stream back pressuring either. https://github.com/Reactive-Extensions/RxJS/issues/508#issuecomment-77513715

briancavalier:
Hey @nackjicholson, thanks for the kind words.

This question has come up a few times.  In my original vision for most.js, I didn't have any plans to deal with node data streams, or any data streams for that matter.  However, it seems like people would find it to be useful, so I'm coming around to being ok with it :)

I've done a bit of investigation, and even [a prototype that uses promises communicated from consumer to producer to indicate demand](https://github.com/cujojs/most/blob/node-read-adapter/lib/source/fromReadableStream.js) (basically as a "stop producing until this promise fulfills" signal).  The prototype works well, but I didn't push it too far ... propagating promises in the reverse direction would get tricky in certain combinators.

I think there needs to be more investigation into other ways to communicate demand.  For example, I could see packaging up the `sink` and `scheduler` arguments into a `subscription` object which provides a communication channel from consumer to producer.  That `subscription` could provide an API similar to node's pause/resume, or something more sophisticated like the `request(N)`/`cancel()` API specified in the [reactive-streams spec](https://github.com/reactive-streams/reactive-streams-jvm#3-subscription-code).

[...]

PR #98, opened 2015-03-10
Start a guide for contributing third-party extensions

unscriptable:
1. Cut from #93
2. Paste to new document
3. Profit

### DIFF
--- a/docs/extending.md
+++ b/docs/extending.md
@@ -1,0 +1,19 @@
+**Stream**
+
+Stream is just a wrapper around a source, and provides a prototype on which public-facing APIs live.
+
+**Source**
+
+A source represents a view of events over time.  It has a single method, `run(sink, scheduler)` which arranges to propagate events to the provided `sink`.  If it needs to deal with time, it can use `scheduler`, which has methods for knowing the current time, and scheduling future tasks in a more performant way than using `setTimeout` directly.
+
+A source's `run` method must return an object with a `dispose` method--similar in concept to RxJS's `Disposable`.
+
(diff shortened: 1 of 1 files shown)


unscriptable:
What needs to be added to make this a real thing?

briancavalier:
This is a great idea!  A few things come to mind:

1. Showing actual (pseudo)code of the Source, Sink, and Disposable interfaces.  I'm not sure what the best format for this would be, though.
1. Providing more guidance on writing a new combinator:
    1. The first choice should always be to build a new combinator using existing ones.  You can do a *lot* with map, filter, flatMap, loop (or scan), sample, and combine.
    1. If it's not possible, or prohibitive (ie too much capturing required, etc.) to build from existing, then create a new Source+Sink.
1. Linking to a couple simple, existing combinators as examples, map and filter come to mind because they are dead simple.

Other ideas?

briancavalier:
Another:

1. Showing an example of a producer Source, and maybe linking to an existing one.

nackjicholson:
Excellent! Don't know of many open source projects that have documented explanations of their internal architecture to help new contributors. I really like the idea. :+1:

briancavalier:
@unscriptable Do you want to work more on this?  Or did you want to hand it off?

briancavalier:
I'd like to get an initiation version of this merged.  Do you think the right name for this document is "extending", or "architecture"?  Maybe "extending" with an "Architecture" section?

briancavalier:
@unscriptable Is it cool with you if I commandeer this and combine it with info in #191?

briancavalier:
Ok, new plan: gonna move the info here and in #191 to the wiki.

Issue #100, opened 2015-03-10
most.unfold does not emit tuple.value that is returned with tuple.done == true


maciasello:
Reproducing code (prints only 0 and 1):
```javascript
most.unfold(function(s) {
    return {value: s, seed: s+1, done: s==2};
}, 0)
.forEach(console.log.bind(console));
```

Additional code :) (prints 0, 1, 2 as the last tuple.value goes as forEach's promise resolution value):
```javascript
most.unfold(function(s) {
    return {value: s, seed: s+1, done: s==2};
}, 0)
.forEach(console.log.bind(console))
.then(console.log.bind(console));
```

There is a question if it is a matter of unclear documentation or a bug.

My first approach was that tuple.value returned in last loop (tuple.done == true) will be emitted as last event in the stream. Instead it is attached to the end event (and is a resolution value of the promise returned by forEach - did not check how other consumers handle it). The docs do not mention neither a concept of a value attached to the end event nor what will be resolution value on forEach. So if it is not a bug then these docs should be updated too.

I would go with a PR but some discussion is needed on the approach to take. My gut feeling is that it should be last "normal" event. Even then some hints about forEach's promise resolution value would be valuable, but I'm not sure what is its purpose now.

davidchase:
I would say its not a bug but expected, since the `unfold` will terminate as soon as `2 == 2`, rather than something like `s > 2` where you will get `0,1,2` because `s` will increment once again before terminating since `2 > 2` is false.

briancavalier:
@maciasello Thanks for raising this issue!  As @davidchase says, the current behavior is correct: `unfold` behaves like "up to" rather than "up to and including".  The bug is the fact that the documentation is lacking!

Note that it may be clearer in some cases to use `take` to declaratively limit the `unfold`, rather than worrying about `done`.  IOW, sometimes it's easier to write an infinite computation and use `take`, and sometimes it's easier to write a finite one that limits itself using `done` :)

briancavalier:
@maciasello See #104 :)

maciasello:
@davidchase, @briancavalier, I understand that it behaves like this right now and the code looks like it was expected behaviour (not a sneaky bug). But the question here is more of what is more natural. In my case it was natural that in the last iteration I had last value to be emitted normally and also an information that it is last one. Do you have any hints why current approach is more natural/typical?

[...]

PR #104, opened 2015-03-11
Clarify how tuple.value is used when tuple.done == true

briancavalier:
This wasn't documented.  See #100


maciasello:
I've added additional info on end signal value. See #105. This changeset is rebased there too, so if #105 is fine this one should be rejected.

briancavalier:
Thanks @maciasello. Closing in favor of #105

Issue #90, opened 2015-03-03
Transducers


jeffrose:
FYI, looks like both Kefir and Bacon are adding support for transducers based on the potential performance gains.

https://github.com/baconjs/bacon.js/issues/444

Apparently Rx already had support.

https://github.com/Reactive-Extensions/RxJS/blob/master/doc/gettingstarted/transducers.md

Just wanted to put this on your radar if it wasn't already.

briancavalier:
Hey @jeffrose!  Yeah, transducers are interesting, I've been following them for a while.

It would be interesting to try an experiment to see if there's a performance gain to be had.  most.js is already [extremely fast](https://github.com/cujojs/most/blob/master/test/perf/README.md), and so I'm not sure if we would see an improvement.  I'm certainly open to trying it, though!

Compatibility with other transducer libs, and presenting a familiar API (to folks familiar with transducers that is) might be other reasons to support transducers, even if there is no perf gain.

briancavalier:
I added experimental transducer support in the [add-transduce branch](https://github.com/cujojs/most/tree/add-transduce).  [The code](https://github.com/cujojs/most/blob/add-transduce/lib/combinator/transduce.js) is pleasantly simple :)

Here are some initial perf runs using [cognitect-labs/transducers-js](https://github.com/cognitect-labs/transducers-js).  I also tested with [jlongster/transducers.js](https://github.com/jlongster/transducers.js) and the perf was nearly identical.

```
filter -> map -> reduce 1000000 integers
-------------------------------------------------------
most        115.79 op/s ± 16.50%   (36 samples)
most transduce   87.74 op/s ±  5.26%   (76 samples)
```

```
filter -> map -> map -> reduce 1000000 integers
-------------------------------------------------------
most         81.69 op/s ±  0.94%   (73 samples)
most transduce   72.81 op/s ±  0.94%   (78 samples)
```

```
filter -> filter -> map -> map -> reduce 1000000 integers
-------------------------------------------------------
most         56.69 op/s ±  0.87%   (80 samples)
most transduce   43.31 op/s ±  5.35%   (64 samples)
```

The performance is good, although most's native combinators are still faster, at least `filter` and `map`.  It's probably worthwhile to test some other combinators to get a broader sample.  Regardless, I think the question is: Given that performance probably *isn't* a benefit, are there other benefits of providing a transducer API that are worth the extra code?

[...]

PR #109, opened 2015-03-16
Add transduce

briancavalier:
Add `transduce` API for using composed transducers with most.js streams.

Still needs:

- [x] Unit tests
- [x] Docs

See #90

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -37,6 +37,8 @@ most.js API
 	* [filter](#filter)
 	* [distinct](#distinct)
 	* [distinctBy](#distinctby)
+1. Transducer support
+	* [transduce](#transduce)
 1. Slicing streams
 	* [slice](#slice)
 	* [take](#take)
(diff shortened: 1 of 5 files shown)


briancavalier:
@jeffrose @kevinbeaty Code, tests, and docs are done!  I'd love to get your feedback before merging :)

briancavalier:
Squashed commits for easier review

lib/combinator/transduce.js
```diff
@@ -0,0 +1,57 @@
...
+
+TransduceSink.prototype.event = function(t, x) {
+	var next = this.xf.step(t, x);
+	if(next != null && next.__transducers_reduced__) {
+		this.sink.end(t, next.value);
+	}
+};
+
+TransduceSink.prototype.end = function(t /*, x */) {
+	this.xf.result(t);
```
kevinbeaty: Is `x` here a possible additional item?  If so, it is valid to step one more time before calling result.

docs/api.md
```diff
@@ -793,6 +795,37 @@ The `equals` function should accept two values and return truthy if the two valu
...
+## Transducer support
+
+### transduce
+
+####`stream.transduce(transducer) -> Stream`
+####`most.transduce(transducer, stream) -> Stream`
+
+Create a new stream by passing items through the provided transducer.
+
+[Transducers](http://simplectic.com/blog/2014/transducers-explained-1/) are composable transformations.  They may map, filter, add items to, drop items from, or otherwise transform an event stream.  The primary benefit of transducers is that they are composable and reusable across any data structures that support them (see note on performance below)
```
kevinbeaty: Thanks for the shout out!

kevinbeaty:
LGTM overall. Exciting stuff

docs/api.md
briancavalier: It's a great article :+1:

lib/combinator/transduce.js
briancavalier: Great question.  Most.js separates the notion of values "in the stream" and an "end signal value".  This is similar, but not identical, to who ES6 generates subtle separate `yield`ed values (which always have `done === true`) from a final `return`ed value (which has `done === false`).

So, now you've got me thinking that a better approach would be to wrap up `t` and `x` into a `{ time, value }` object and do `this.xf.result({ time: t, value: x })`, and then `Transformer.prototype.result`, could do `this.sink.end(time, value)`.  That would propagate this final value as the "end signal value".

briancavalier:
Thanks @kevinbeaty, really appreciate the thorough review!

kevinbeaty:
No problem.  

Just a heads up: The property that signals early termination `__transducers_reduced__` may be changing (for the better): cognitect-labs/transducers-js#20 . 

Not sure if that means you want to wait on this, or support the new version in the future.

briancavalier:
Thanks for the heads up.  Any idea what the ETA is on the switchover to the new protocol?  If it's far enough out, I don't mind merging this asap and then updating when the new protocol lands.  If it's "close", then it may be better to wait.

kevinbeaty:
Hard to say as there are multiple libraries involved. The discussion only started today but it seemed to reach a good common ground quickly. I don't think there is any harm in switching over later.  The current versions of the libraries follow the `__transducers_reduced__` convention that you implemented here.

briancavalier:
Cool.  The only snag I can see on switching later is someone using an older transducer lib after most.js switches to support the newer protocol.  I think we can mitigate that via semver.  For example, I can release a new major (or minor if we're still < 1.0) version when most.js switches protocols.

So, yep, I'm cool with getting this merged asap and switching later.

briancavalier:
Whew, ok, after a lot of thought and experimentation with different ways to thread *both* the event time and value through a transducer, I have it working.  Unfortunately, it required a stateful solution, but that's ok since it's hidden inside the machinery.  I also added a unit test for the scenario where a stream combinator ends the stream early (rather than a transducer combinator), to ensure that didn't wreak havoc on transducers at a later point in the stream.

I think this is ready to land!

briancavalier:
Squashed again

briancavalier:
Added jsdoc and squashed.  I'll merge this later today unless anyone has more suggestions.

briancavalier:
:boom:

PR #112, opened 2015-03-19
Add wire plugin to allow stream creation directly from wire specs

maciasello:
I'm opening the PR mainly for discussion.

Now it supports three different ways to create streams. It may be misleading and a decision to get rid of some could be taken (@briancavalier ?). It may be also that other ways are found suitable.

Docs need improvement in particular types reference. I'm also open to changing parameter names. Esp. `inputStream` is not the best choice IMO.

Types involving passing functions are omitted now (no problem here - just being a bit lazy :D).

I'm thinking also to provide helpers such as being able to `merge` an array of streams (I know it could be achieved by `from` and `join` but maybe this would be helpful).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ It features an ultra-high performance, low overhead architecture, APIs for easil
 ## Learn more
 
 * [API docs](docs/api.md)
+* [wire.js plugin](docs/wire.md)
 * [Examples](examples)
 * [Get it](#get-it)
 
--- a/docs/wire.md
+++ b/docs/wire.md
@@ -1,21 +1,21 @@
 # Table of contents
 
 1. Introduction
-1. ['stream' wire plugin](#module-most-wire)
-	1. ['stream' Factory](#wire-stream-factory)
-	1. ['stream' resolver](#wire-stream-resolver)
-1. ['stream' components creation API](#stream-components-creation-api)
+1. [`stream` wire plugin](#module-most-wire)
+	1. [`stream` Factory](#wire-stream-factory)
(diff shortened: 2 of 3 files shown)


briancavalier:
Hey @maciasello, this is a very interesting idea.  I'll post more thoughts a bit later today after I have a chance to absorb it :)

briancavalier:
@maciasello A couple of initial thoughts

The question of where wire plugins should live has come up a good bit.  For example, should a most-wire plugin live in the wire.js repo or the most.js repo.  There's a third option, and I've come to think it's the best one: it should live in its own repo.  So, how do you feel about moving this to its own repo?  It makes it clear that it's your project (i.e. you get credit), and it allows people to `npm install` it directly.

It's probably already possible to create streams using @foxdonut's [recent pull-request](https://github.com/cujojs/wire/pull/170), which allows calling a function like `most.period` directly via wire's `create`.

That said, one thing I think would be very interesting to think about is how to create streams *from* other things in a wire spec.  For example, could wire allow people to easily inject a stream for a DOM events?  That might be similar to your idea for using a `$ref`.  Another thought: could a wire plugin allow people to easily connect things with streams.  The AOP connections are already similar, but much less powerful than in some ways than a full stream.  For example, you can't filter AOP connections.  I haven't thought through any of that to know how it might work or what it might look like in a wire spec, but just wanted to throw it all out there to see if it generates any ideas.

briancavalier:
@maciasello Are you interested in continuing this?

briancavalier:
Closing, but please feel free to reopen.

PR #113, opened 2015-03-23
1st pass at recipes

davidchase:
@briancavalier Sorry for the delay finally got my computer back.. i've included two recipes as 1st pass at it please let me know your thoughts or if I am way off, etc :smile:

### DIFF
--- a/docs/recipes/event-delegation.md
+++ b/docs/recipes/event-delegation.md
@@ -1,0 +1,45 @@
+# Event delegation 
+
+In the 1st example we use `most.fromEvent` to create a stream from
+a `click` event. 
+
+By adding `filter` in conjunction with [element matches](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches)
+we can specify which element we want to observe and in this case get the `textContent`
+from the element.
+
+The 2nd example is similar, however we use jQuery to handle event delegation.
+
+We use `most.create` to produce a push-stream off of those click events.
+Essentially each time a user clicks on the `.inner` that event is added to the
+stream which can be observed and processed in a similar manner to the 1st example.
+
+It should be noted that the `dispose` function which calls `container.off` only happens
+if the implementor would call `end()`, otherwise the stream will not end and
+the event will still be bound.
+
+```js
+var most = require('most');
+
+var container = document.querySelector('.container');
+most.fromEvent('click', container)
+    .filter(function(e){
+        return e.target.matches('.hello');
+    })
+    .observe(function(e){
+        return console.log(e.target.textContent)
+    });
+```
+
+```js
+var $ = require('jquery');
+var container = $(".parent");
+most.create(function (add) {
+    container.on("click", '.inner', add);
+    return function () {
+        return container.off("click",'.inner', add);
+    };
+})
+.observe(function(e){
+    return console.log(e.target.textContent)
+});
+```
--- a/docs/recipes/event-delegation.md
+++ b/docs/recipes/event-delegation.md
@@ -1,6 +1,6 @@
 # Event delegation 
 
-In the 1st example we use `most.fromEvent` to create a stream from
+In the 1st example we use [most.fromEvent](../api.md#mostfromevent) to create a stream from
 a `click` event. 
 
 By adding `filter` in conjunction with [element matches](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches)
@@ -9,13 +9,11 @@ from the element.
 
 The 2nd example is similar, however we use jQuery to handle event delegation.
 
-We use `most.create` to produce a push-stream off of those click events.
+We use [most.create](../api.md#mostcreate) to produce a push-stream off of those click events.
 Essentially each time a user clicks on the `.inner` that event is added to the
 stream which can be observed and processed in a similar manner to the 1st example.
 
-It should be noted that the `dispose` function which calls `container.off` only happens
-if the implementor would call `end()`, otherwise the stream will not end and
-the event will still be bound.
+
 
 ```js
 var most = require('most');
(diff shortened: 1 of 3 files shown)


briancavalier:
@davidchase Woot! I think you picked two really good recipes to start with.  I especially like that you showed both the straight `most.fromEvent` and jQuery versions of the event delegation recipe.

I thought of a few questions while reading over them, and want to see what you think.  It'd be great to hear from @unscriptable on these, too.

1. Should recipes be "runnable"?  This may put them into the same category as the stuff the current examples dir, but maybe not.  Some variations on what "runnable" might mean:
  * Literally runnable: each recipe is implemented as a runnable nodejs program, or a single-page html "app".  This requires more infrastructure in each recipe.
  * Cut-and-pasteable: the code shown in each recipe shows a complete example (including `require`s, for example) that can be pasted and run with minimal (but non-zero) effort.
2. Slightly related: Should we show code in ES6?  This generally makes "runnable" more difficult, but I like the idea of helping people get accustomed to seeing ES6.

davidchase:
@briancavalier you know whats interesting i tested out these recipes in http://requirebin.com/ which was quite joyful :smile: i played around with most, and native fetch and even tried out the polyfill for fetch..

briancavalier:
@davidchase That's pretty awesome.  If they're already paste-able into requirebin, then I vote for sticking with that as our approach!  Maybe we just add a note to recipes/README.md that tells people they can do that, and links to requirebin.  What do you think?

davidchase:
yeah i think thats a great idea, people should be able to tinker with the recipes before using them in their projects. will add the link to requirebin in the readme and update it will the other recipe as well(oops).
only caveat is that saving in requirebin saves to github gists, and as of right now i cant event save a file... maybe you can try or we can add links later

briancavalier:
Looking good.  Are there other recipes you'd like to add in this PR?  Or, would you prefer to get this PR merged asap as a baseline, and we can add more recipes in upcoming PRs?

davidchase:
I don't mind if we merge them as a baseline, that way we can get the ball rolling, maybe have other people like yourself, etc start contributing to the recipes "pot"... what are your thoughts?

briancavalier:
Sounds good to me.  I'll give this one final review later tonight, and if all looks good, merge it.

docs/recipes/event-delegation.md
```diff
@@ -0,0 +1,45 @@
...
+from the element.
+
+The 2nd example is similar, however we use jQuery to handle event delegation.
+
+We use `most.create` to produce a push-stream off of those click events.
+Essentially each time a user clicks on the `.inner` that event is added to the
+stream which can be observed and processed in a similar manner to the 1st example.
+
+It should be noted that the `dispose` function which calls `container.off` only happens
+if the implementor would call `end()`, otherwise the stream will not end and
```
briancavalier: Technically, the `dispose` will be called when there's no *demand* for events.  That can happen in several ways.  One way is when the stream producer (i.e. the function passed to `most.create`) explicitly calls `end`.  Another is when all consumers have "lost interest".  For example, if 2 callers start consuming the stream, and the first uses `.take(5)` and the second uses `.take(10)`, the first 5 events will be delivered to both consumers.  Then, the first loses interest (because of `take(5)`).  Since the second consumer is still interested, `dispose` is not called yet.  The second consumer receives 5 more events, then loses interest (because of `take(10)`).  At that point, `dispose` is called because there is no longer any demand for events.

That's probably waaaayyyyy to long-winded to explain in the recipe :)  So, one option would be to link to the [docs for `most.create`](https://github.com/cujojs/most/blob/master/docs/api.md#mostcreate) which explain how it works.

docs/recipes/event-delegation.md
```diff
@@ -0,0 +1,45 @@
...
+        return console.log(e.target.textContent)
+    });
+```
+
+```js
+var $ = require('jquery');
+var container = $(".parent");
+most.create(function (add) {
+    container.on("click", '.inner', add);
+    return function () {
```
briancavalier: Let's name this function `dispose` to help correlate with the text above.

docs/recipes/event-delegation.md
```diff
@@ -0,0 +1,45 @@
+# Event delegation 
+
+In the 1st example we use `most.fromEvent` to create a stream from
```
briancavalier: Link to `most.fromEvent` api docs

briancavalier: In general, I think we should link to the docs for the first occurrence of every API function mentioned in the recipe.  What do you think?

docs/recipes/unfold-promises.md
```diff
@@ -0,0 +1,37 @@
...
+```js
+var most = require('most');
+var urls = ['http://reqr.es/api/users?page=2', 
+            'http://reqr.es/api/users?delay=3', 
+            'http://reqr.es/api/users?page=3'];
+
+most.unfold(function (urls) {
+    return urls.length === 0 ? {
+        done: true
+    } : fetch(urls[0]).then(function(content){
```
briancavalier: This is a tricky thing to format.  Took me a minute to parse it.  Maybe something like:

```js
return urls.length === 0
    ? { done: true }
    : fetch(urls[0]).then(function(content) {
```

I dunno, what do you think?

davidchase: Sorry about that, & I like your style :)

docs/recipes/event-delegation.md
davidchase: so should i just remove my note about the dispose and link towards `most.create`?

briancavalier: Sounds great!

davidchase:
@briancavalier made the updates per request, let me know :+1:

briancavalier:
Looks great, @davidchase.  I'm about to release 0.13, and it'll really nice to have this in the release.

briancavalier:
@davidchase [here we go](https://github.com/cujojs/most#learn-more)

PR #118, opened 2015-04-10
Add ES7 Observable strawman

briancavalier:
For Discussion.

This provides an ES7 Observable(producer) constructor, as most.Observable, as well as `observable['@@observer']` and observable.observer methods for observing.

See:

* https://github.com/jhusain/asyncgenerator
* https://github.com/cujojs/most/pull/72#issuecomment-91162216

Example:

```js
var Observable = require('most').Observable;

var observable = new Observable(function(generator) {
	console.log(1);
	generator.next(2);
});

observable['@@observer']({
	next: function (value) {
		console.log(value);
	}
});

console.log(3);

// Logs 3 1 2
```

### DIFF
--- a/lib/Observable.js
+++ b/lib/Observable.js
@@ -1,0 +1,32 @@
+var Stream = require('./Stream');
+var ObservableSource = require('./source/ObservableSource');
+var MulticastSource = require('./source/MulticastSource');
+var AsyncSource = require('./source/AsyncSource');
+var scheduler = require('./scheduler/defaultScheduler');
+
+exports.create = createObservable;
+exports.observer = observer;
+
+function createObservable(generator) {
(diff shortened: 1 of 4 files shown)


briancavalier:
I can think of 3 paths we can take:

1. Don't merge this.  Let the es7 observable direction play out, or
2. Change Publisher API to match ES generator API (i.e. `next`, `return`, and `throw`) and merge, or
3. Merge as is

I'm definitely seeking feedback.

I don't want to create a new API that becomes obsolete if react is already planning to support an es7 observable style API.  However, if that is going to take a while to be practically usable, it may still make sense to release Publisher.

Thoughts?

briancavalier:
This PR is far out of date with the latest es7.  It's still on the roadmap (see #180).

PR #120, opened 2015-04-11
Simplify switch() using until() and concurrentMergeMap()

briancavalier:
This reuses `until` and `concurrentMergeMap` to achieve the same behavior with much less code.


briancavalier:
Splitting the `map` out of concurrentMergeMap has the interesting side effect of allowing the map step to be gobbled up by stream fusion, like any other `map` call.

PR #121, opened 2015-04-13
Fix debounce final event handling

briancavalier:
When debounce ends, if there is a pending event, it is now emitted before ending the stream.  This commit also improves the reliability of several time-based tests, including debounce tests.


unscriptable:
:+1:

PR #126, opened 2015-04-20
Add currying

briancavalier:
Add curry and compose, curry all public API functions.
    
Both exact 2/3-ary functions *and* varargs functions are curried. Varargs functions are curried to a minimum number of args.  For example, since it doesn't make much sense to call combine with 2 args, most.combine is curried to 3 arg-minimum.  So, most.combine(f) returns combine curried with f, and most.combine(f, a) returns combine curried with f and a.  However, most.combine(f, a, b) calls combine, as does most.combine(f, a, b, c) etc.
    
That's totally up for discussion!  Another perfectly reasonable approach would be not to curry varargs functions at all.

### DIFF
--- a/lib/compose.js
+++ b/lib/compose.js
@@ -1,0 +1,56 @@
+/** @license MIT License (c) copyright 2010-2015 original author or authors */
+/** @author Brian Cavalier */
+/** @author John Hann */
+
+exports.compose = compose;
+exports.pipeline = pipeline;
+
+function compose(g, f /*, ... */) {
+	var l = arguments.length;
+
+    if(l === 1) {
+		return g;
+	}
+
+	if(l === 2) {
+		return function(x) {
+			return g(f(x));
+		};
+	}
+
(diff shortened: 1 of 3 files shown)


briancavalier:
There is at least one potential blocker: `most.periodic` has an optional argument to specify the value.  If you don't specify, it uses `undefined`.  But if we curry `most.periodic` it will mean that not supplying a value (i.e. only a time) lead to partial application.  That's a breaking change.  We could go back to `most.periodic` not accepting a value, and forcing callers to use `.map` or `.constant` to provide the value.  We could also not curry `most.periodic`, but I'd rather be consistent if we can.

So, I'm really not sure what's best.  Any thoughts?

unscriptable:
You've convinced me that currying might be better in a future release. Lots to consider.

briancavalier:
Closing this in favor of putting it on the roadmap and getting more feedback.

Issue #96, opened 2015-03-08
Question about comparing most & bacon.js


aaronshaf:
Brian, why cujo? What makes it special over, say, bacon.js and others?

Thanks :-)

briancavalier:
@aaronshaf I'll try to describe some of the motivations that led to most.js and some of the things we think are interesting about it.

I'll be up front by saying that I don't really think of most.js, bacon, rxjs, kefir, etc. as a competition that most.js needs to "win".  They are all useful tools in the reactive programming space.  All of them have likely learned from and borrowed ideas from each other.  The community will benefit from friendly competition *and* cooperation.

I'll be honest about my criticism of those libs here, but I have a ton of respect for them and their authors & contributors.  I'm sure they have criticisms of most.js they could share.

In terms of personal motivation wrt reactive programming, I think it's an important concept that provides a lot of advantages: it naturally handles asynchrony and provides a model for dealing with complex data *and time* flow while also removing (or at least lessening) the need to resort to shared mutable state.  It has many applications: interactive UIs and animation, client-server communication, IoT and sensor networks, etc.

In terms of motivation for most.js, I looked at existing implementations and felt there were several areas that could be improved upon. Here are a few of them:

**Performance**

Looking at the architecture and implementation of other libs, I felt it should be possible to get at least 1 to 2 orders of magnitude better performance.  Kefir had done some good work in this area, but I felt that there were still plenty of perf gains to be had.

The [perf test results](https://github.com/cujojs/most/tree/master/test/perf) indicates that most.js is achieving its goal in this area.  My hope is that by publishing those numbers, and showing what is possible, other libs will improve as well.

I've only measured memory informally, so I don't want to comment on it in detail.  However, it is also favorable compared to other libs.

**Modularity**

Other libs seemed not to have truly modular architectures.  Bacon is implemented in a single file.  RxJS is only "modular" in the sense that it's files must be concatenated in the correct order. I feel Kefir is a bit more modular, but as far as I could tell, also did not embrace a real module system such as CommonJS.

Most.js it highly modular CommonJS (I wanted to go ES6, but felt it still required a bit too much tooling, and that generated code would likely impact performance goals).  It's internal Stream/Source/Sink architecture and APIs are simple, concise, and well defined (although not yet documented, which is an obvious problem!).  Combinators can be implemented entirely in terms of that API.  They don't need to use any private details.  This makes it easy to implement new combinators externally (ie in contrib repos, for example) while also guaranteeing they can still be high performance.

[...]

PR #131, opened 2015-04-23
Update README with info from #96

### LINKED ISSUE #99: Update README with info from #96
Some of the info in #96 may be helpful to others, and would be easier to find if it lived in the README.

briancavalier:
Extract some info from #96 and integrate into README.

Close #99


TylorS:
:+1:

PR #135, opened 2015-05-09
Use `lodash` functions instead of those in `base.js`

izaakschroeder:
Still a WIP, just opening for discussion. Not entirely sure why CI is failing yet.

Sometimes it is helpful to re-invent the wheel, but when you have a fast, battle-tested library at your disposal there's no reason not to leverage it! A leaner library is always a good thing.

Advantages:
 * Variadic functions can now accept arrays as arguments, so `(foo)` is as valid as `([foo, foo])` or `([foo, [foo], foo], [foo])`,
 * Less code to maintain, debug and test,
 * Don't have to worry about browser incompatibility for those functions,
 * More concise code (e.g. one `pluck` instead of needing a whole `getSource` function),
 * People who use `lodash` get smaller builds when they use tools like `webpack` since the common code can be shared,
 * Performance seems to be maintained.

Disadvantages:
 * Will probably increase build size slightly for standalone `most`; this can probably be addressed with some tweaking.

Other thoughts:
 * Function order is "backwards"; using `lodash-fp` instead "fixes" this, but I imagine most implementors won't be using the fp-specific library; doing this removes the advantage of shared code.
 * Being able to pick your desired Promise library would also help cut down on the build size for those already using Promise libraries and allow for better code sharing (`bluebird` for days boys). This could be achieved with `webpack` via something like `if (__PROMISE__) { module.exports = require(__PROMISE__); } else { module.exports = require('when'); }` About 25KB (~40%) of `most` size comes from `when`.

Performance sample:
```
master:
concatMap 1000 x 1000 streams
most        121.00 op/s ±  4.72%   (73 samples)
rx            1.12 op/s ± 15.72%    (7 samples)
kefir        11.49 op/s ±  5.40%   (31 samples)
bacon         0.95 op/s ±  1.94%    (7 samples)
lodash        5.05 op/s ±  8.85%   (17 samples)
Array         0.33 op/s ±  2.17%    (5 samples)

baseless:
concatMap 1000 x 1000 streams
most        120.51 op/s ±  3.94%   (77 samples)
rx            1.33 op/s ±  5.07%    (8 samples)
kefir        11.12 op/s ±  4.58%   (30 samples)
bacon         0.92 op/s ±  8.09%    (7 samples)
lodash        4.99 op/s ±  9.17%   (17 samples)
Array         0.26 op/s ± 48.56%    (5 samples)
```

Build size results:

```
master:
webpack -p --entry ./most --target web most.min.js
du -hs most.min.js
60K    most.min.js

baseless:
webpack -p --entry ./most --target web most.min.js
du -hs most.min.js
76K    most.min.js
```


briancavalier:
Hey @izaakschroeder, thanks for proposing this.  It's impressive that you've dug deeply enough into the most.js code to figure this out :)  I have a huge amount of respect for lodash and its developers.

A goal of most.js is to have zero dependencies.  Right now, we have a dependency on `when` simply because it hasn't been practical to depend on native `Promise`, and we felt that "bring your own promise" was still a little too much to ask.  We wanted the best possible out-of-the-box experience.  The native Promise landscape has improved, so we'll probably re-evaluate going with "bring your own promise" after 1.0.0 (0.14.0 is the 1.0.0 release candidate).

Our API design philosophy differs somewhat from that of lodash.  We prefer functions that do exactly one thing, and accept more tightly focused input types rather than accepting many possible input types.  Accepting variadic or array doesn't fit with out API design philosophy.  However, with "functions first, data last" argument ordering, and [ES6 spread args](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator), it's a moot point: e.g. `most.combine(f, ...streams);` or `most.combine(f, stream1, stream2, stream3);`

The base module has proven to be fast and solid, and contains only what most.js needs.  It also fits with our zero dependencies goal.  We'll stick with it for the foreseeable future.

PR #138, opened 2015-05-20
Introduce TestScheduler and improve unit tests

briancavalier:
This adds an internal API for running streams on a particular scheduler, as well as a TestScheduler that runs streams on a virtual timeline.  Because it doesn't have to map onto real-world time, it gets rid of nondeterminism in some tests due to timer resolution, and speeds up unit tests a good bit.  They're over 2x faster now (see `total` in each below):

master:

```
> time ./node_modules/.bin/buster-test -e node
141 tests, 279 assertions, 1 runtime ... OK
./node_modules/.bin/buster-test -e node  0.95s user 0.08s system 45% cpu 2.256 total
```

introduce-test-scheduler branch:

```
>time ./node_modules/.bin/buster-test -e node
141 tests, 295 assertions, 1 runtime ... OK
./node_modules/.bin/buster-test -e node  0.89s user 0.08s system 89% cpu 1.076 total
```

The scheduler API still isn't considered public, but this PR is a small step in that direction.


lib/scheduler/Scheduler.js
```diff
@@ -118,6 +116,7 @@ Scheduler.prototype._runReadyTasks = function() {
 	// TODO: Consider using findInsertion instead of linear scan
 	for(i=0; i<l; ++i) {
 		task = tasks[i];
+		//console.log(task.time, now);
```
briancavalier: Oops, lots of leftover `console.log` in here.  Will remove

test/build-test.js
```diff
@@ -67,11 +68,16 @@ describe('build', function() {
 			var s1 = delay(1, fromArray([1,2]));
 			var s2 = fromArray([3,4]);
 
-			return reduce(function(a, x) {
-				return a.concat(x);
-			}, [], build.concat(s1, s2)).then(function(a) {
-				expect(a).toEqual([1,2,3,4]);
-			});
+			var scheduler = new TestScheduler();
+			scheduler.tick(1);
```
briancavalier: This is the gist of the TestScheduler.  Create one, and then call `tick` to arrange for it to run virtual timeline of a certain number of ticks.

test/build-test.js
```diff
@@ -67,11 +68,16 @@ describe('build', function() {
...
+
+			return scheduler.collect(build.concat(s1, s2))
```
briancavalier: Then, call `collect` to run a stream on the virtual timeline and collect all the events as `{ time, value }` pairs.

Issue #139, opened 2015-05-31
The mapping function passed to `map` is invoked once for every observer for one event on base stream


maciasello:
The simplest reproducing scenario:
```js
var orig;
orig = most.of(7).map(function() {
  console.log("expensive operation");
  return 5;
});
orig.observe(function(e) {
  console.log("obs 1", e);
});
orig.observe(function(e) {
  console.log("obs 2", e);
});
```

My usage is that for every user event (think "click") there is an expensive operation performed (e.g. AJAX call for to-do list) then multiple derivative streams carrying the result processed in different ways (e.g. a stream with to-do list content and a stream with to-do list length) are created. Each of the derivative streams is observed by one observer, but still the expensive operation is invoked many times - once for each observer even though there was one "click" only.

A bit more complex scenario (to ilustrate the derivative streams) - it bahaves the same if `.fromEvent(...)` is changed to `.of(...)`:
```js
var derivative1, derivative2, orig;

orig = most.fromEvent("click", someDiv).map(function(e) {
  console.log("expensive operation");
  return 7;
});

derivative1 = orig.map(function(e) {
  console.log("sub-op 1");
  return e + 1;
});

derivative1.observe(function(e) {
  console.log("observer 1", e);
});

derivative2 = orig.map(function(e) {
  console.log("sub-op 2");
  return e - 1;
});

derivative2.observe(function(e) {
  console.log("observer 2", e);
});
```
Console output on the latter:
```
expensive operation
sub-op 1
observer 1 8
expensive operation
sub-op 2
observer 2 6
```

Is this expected or a bug? While I can [partially] visualize to myself the multiple sinks pulling the events from sources and thus causing multiple invocations of a mapper it is against my gut feeling that one source event should be mapped once and then fed to different consumers. Is this reasoning correct?

briancavalier:
Hey @maciasello!  I'll call this known, but in search of the best solution.  Because most pre-1.0 has been heavily focused on performance, it optimizes for the single-subscriber case by not bothering to maintain subscriber lists at each node.  Multicast subscriptions can be added easily [via composition](https://github.com/cujojs/most/blob/master/lib/source/MulticastSource.js).  However, we haven't exposed an API for it yet.

One issue is that there's no way for each combinator to know if the stream it returns will end up having multiple observers.  I think there are 2 possible solutions to consider for 1.0:

[...]

PR #143, opened 2015-06-10
Add multicast combinator

maciasello:
This adds stream.multicast() that converts a stream into a multicast stream (as one of the solutions to #139).

This will probably be superseded by internal changes that will handle multicasting in a more user-friendly way. Maybe it should be denoted as deprecated from the very beginning? ;)

No documentation (yet). No tests as it feels like too simple to break/test.

I've added new combinator even though it is extremely simple (could be done in most.js main file). Any hints if this is a good approach?

### DIFF
--- a/lib/combinator/multicast.js
+++ b/lib/combinator/multicast.js
@@ -1,6 +1,7 @@
 /** @license MIT License (c) copyright 2010-2015 original author or authors */
 /** @author Brian Cavalier */
 /** @author John Hann */
+/** @contributor Maciej Ligenza */
 
 var Stream = require('../Stream');
 var MulticastSource = require('../source/MulticastSource');
@@ -9,7 +10,7 @@ exports.multicast = multicast;
 
 /**
  * Transform the stream into multicast stream.  That means that many subscribers
- * to the stream will not cause multiple invokations of the internal machinery.
+ * to the stream will not cause multiple invocations of the internal machinery.
  * @param {Stream} stream to be transformed.
  * @returns {Stream} new stream which will multicast events to all observers.
  */
--- a/lib/combinator/multicast.js
+++ b/lib/combinator/multicast.js
@@ -1,0 +1,18 @@
+/** @license MIT License (c) copyright 2010-2015 original author or authors */
+/** @author Brian Cavalier */
+/** @author John Hann */
+
+var Stream = require('../Stream');
+var MulticastSource = require('../source/MulticastSource');
+
+exports.multicast = multicast;
+
+/**
+ * Transform the stream into multicast stream.  That means that many subscribers
+ * to the stream will not cause multiple invokations of the internal machinery.
+ * @param {Stream} stream to be transformed.
+ * @returns {Stream} new stream which will multicast events to all observers.
+ */
+function multicast(stream) {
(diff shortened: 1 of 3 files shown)


briancavalier:
Looks great.

> Maybe it should be denoted as deprecated from the very beginning?

LOL.  I think I'm ok with adding it, and then only deprecating it once we have something we believe is clearly better.

> No tests as it feels like too simple to break/test.

I can think of one simple test that seems worthwhile: Ensuring that a map function preceding the multicast is only called once when there are 2 observers.  That seems like enough.  What do you think?

> Any hints if this is a good approach?

Absolutely it is.  Consistency ftw.

briancavalier:
I've been toying with renaming `MulticastSource` to `SharedSource`.  Any thoughts on that name, and possibly naming this new method `share()` instead of `multicast()`?

lib/combinator/multicast.js
```diff
@@ -0,0 +1,18 @@
+/** @license MIT License (c) copyright 2010-2015 original author or authors */
+/** @author Brian Cavalier */
+/** @author John Hann */
```
briancavalier: Please do add yourself as a `@contributor`.

most.js
```diff
@@ -649,3 +649,20 @@ exports.throwError   = errors.throwError;
 Stream.prototype.flatMapError = function(f) {
 	return errors.flatMapError(f, this);
 };
+
+//-----------------------------------------------------------------------
+// Changing behavior
+
+var multicast = require('./lib/combinator/multicast').multicast;
+
+
```
briancavalier: extra line break

most.js
```diff
@@ -649,3 +649,20 @@ exports.throwError   = errors.throwError;
 Stream.prototype.flatMapError = function(f) {
 	return errors.flatMapError(f, this);
 };
+
+//-----------------------------------------------------------------------
+// Changing behavior
```
briancavalier: How about "Sharing" or "Multicasting"?

most.js
```diff
@@ -649,3 +649,20 @@ exports.throwError   = errors.throwError;
...
+
+var multicast = require('./lib/combinator/multicast').multicast;
+
+
+exports.multicast = multicast;
+
+/**
+ * Transform the stream into multicast stream.  That means that many subscribers
+ * to the stream will not cause multiple invokations of the internal machinery.
```
briancavalier: invokations -> invocations

lib/combinator/multicast.js
```diff
@@ -0,0 +1,18 @@
...
+
+var Stream = require('../Stream');
+var MulticastSource = require('../source/MulticastSource');
+
+exports.multicast = multicast;
+
+/**
+ * Transform the stream into multicast stream.  That means that many subscribers
+ * to the stream will not cause multiple invokations of the internal machinery.
```
briancavalier: invokations -> invocations

briancavalier:
Cool, thanks for the update @maciasello.  Any thoughts on naming (share vs. multicast) and a simple unit test?

maciasello:
As for naming I'm more for multicast than sharing. Simple tests added.

briancavalier:
Awesome, thanks for working on this @maciasello.

PR #144, opened 2015-06-12
Add timeslot ordering

briancavalier:
Tasks can be ordered even within a single timeslot.  That allows actions to be taken "later" in the same timeslot.  For example, combine can behave predictably in the face of simultaneous events:

```js
let x = most.whatever();

// combine x with itself
// Each time an event occurs in x, only *one* event should occur in x2
let x2 = most.combine((xa, xb) => xa + xb, x, x);
```

### DIFF
--- a/lib/combinator/combine.js
+++ b/lib/combinator/combine.js
@@ -5,8 +5,8 @@
 var Stream = require('../Stream');
 var transform = require('./transform');
 var core = require('../source/core');
-var Pipe = require('../sink/Pipe');
 var IndexSink = require('../sink/IndexSink');
+var PropagateTask = require('../scheduler/PropagateTask');
 var CompoundDisposable = require('../disposable/CompoundDisposable');
 var base = require('../base');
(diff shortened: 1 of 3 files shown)


lib/combinator/combine.js
```diff
@@ -67,29 +67,49 @@ Combine.prototype.run = function(sink, scheduler) {
...
+	if(!this.active) {
+		this.active = this.sinks.every(hasValue);
 	}
 
-	if(this.ready) {
-		// TODO: Maybe cache values in their own array once this.ready
-		this.sink.event(t, invoke(this.f, map(getValue, this.sinks)));
+	if(this.active && t > this.time) {
+		this.time = t;
+		this.task = this.scheduler.at(t, new CombineTask(this.f, this.sinks, this.sink), 1);
```
briancavalier: Hardcoding `1` here isn't a good solution.  There should be some constants, or maybe use `scheduler.defaultPriority + 1` or something like that instead.

Issue #146, opened 2015-06-23
Improving the drag'n'drop example


PhiLhoSoft:
I tried the drag'n'drop example https://github.com/cujojs/most/blob/master/examples/drag-n-drop and tried to improve it a bit. Here is my feedback.

I wanted a quick try, without first installing the library. Apparently, there is no way to try most in jsFiddle, Plunker or similar. As advised in the Recipes page, only RequireBin can allow an online example. I think it is a bit unfortunate, as it limits online examples. Stack Overflow answers often show snippets in the above sites, or even inline in the answer.

RequireBin is nice, as it does the job, but it annoying in the sense we cannot debug the code (it is inlined in an URL-encoded script tag!) except by putting console.log everywhere...
Also crashed Firefox with an out of memory error, although I don't know if that's RequireBin fault, Most's fault, or mine... :smile: 

Note that I had to replace the `module.exports` assignment with an IIFE.

Maybe you can produce a distribution / bin version producing a global object (à la jQuery / Underscore / etc.) which can then be used from a CDN.

Anyway, back to the example.

First thing I remark: when I drag the box in Firefox, part of the text is selected.
I fixed that by adding `e.preventDefault();` at the start of drag's map function (line 22).

Second thing: when I start to drag, the box jumps to center itself on the cursor. Awful effect...
I fixed that by removing the offset() function which defines an arbitrary offset, by defining an empty dragOffset object, passed to the `reduce` call, and by filling it in the grab event (after preventDefault) as:

			dragOffset.dx = e.clientX - draggable.offsetLeft;
			dragOffset.dy = e.clientY - draggable.offsetTop;

A classical in D'n'D code.

Full code is as follows:
```
var most = require('most');

var DROP = 0, GRAB = 1, DRAG = 2;

(function()
{
	// The thing we want to make draggable
	var draggable = document.querySelector('.draggable');
	var dragOffset = {};

	// A higher-order stream (stream whose events are themselves streams)
	// A mousedown DOM event generates a stream event which is
	// a stream of 1 GRAB followed by DRAGS (ie mousemoves).
	var drag = most.fromEvent('mousedown', draggable)
		.map(function(e)
		{
			e.preventDefault();
      
			console.log('e: ' + e.clientX + ', ' + e.clientY);
			console.log('draggable: ' + draggable.offsetLeft + ', ' + draggable.offsetTop + " (" + draggable.offsetWidth + "x" + draggable.offsetHeight + ")");
[...]

PR #147, opened 2015-07-01
Improve drag'n'drop example

PhiLhoSoft:
- Mouse move and up events are on the whole page, avoiding loosing the
box when moving fast.
- Memorize the mouse down location, to avoid centering box on mouse.
- Prevent default mouse down event to avoid selecting text in box in FF.

### DIFF
--- a/examples/drag-n-drop/main.js
+++ b/examples/drag-n-drop/main.js
@@ -11,29 +11,39 @@ var most = require('most');
 var DROP = 0, GRAB = 1, DRAG = 2;
 
 module.exports = function() {
+	// The area where we want to do the dragging
+	var area = document.querySelector('.dragging-area');
 	// The thing we want to make draggable
 	var draggable = document.querySelector('.draggable');
+	var dragOffset = {};
 
-	// A higher-order stream (stream whose events are themselves streams)
+	// A higher-order stream (stream whose items are themselves streams)
 	// A mousedown DOM event generates a stream event which is
-	// a stream of 1 GRAB followed by DRAGS (ie mousemoves).
+	// a stream of 1 GRAB followed by DRAGs (ie mousemoves).
 	var drag = most.fromEvent('mousedown', draggable)
 		.map(function(e) {
-			return most.fromEvent('mousemove', e.target)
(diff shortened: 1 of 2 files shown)


PhiLhoSoft:
Corresponding to issue #146.

briancavalier:
Thanks @PhiLhoSoft!  I should have time to look at this in the next couple days.

PhiLhoSoft:
To test it manually (listing the issues I saw in the original) demo):
- In FF, start dragging; the text at the top of the box should not be (partially) selected;
- Click on a corner of the box, start dragging; the box must not recenter on the mouse cursor
- Drag quickly across the page; the box must follow the cursor, not be left behind.

examples/drag-n-drop/main.js
```diff
@@ -11,29 +11,39 @@ var most = require('most');
 var DROP = 0, GRAB = 1, DRAG = 2;
 
 module.exports = function() {
+	// The area where we want to do the dragging
+	var area = document.querySelector('.dragging-area');
 	// The thing we want to make draggable
 	var draggable = document.querySelector('.draggable');
+	var dragOffset = {};
```
briancavalier: Ah, interesting.  I wonder if there's a way to get the offset right without having to capture and update this as shared mutable state.  For example, could it also pass through the stream, perhaps as a part of the dragInfo object that's already flowing through the stream?

PhiLhoSoft: Yes, I wasn't much happy with that, but it was the quick easy way to fix the example without changing too much its structure.
Somehow, the reduce call used for its side effect is a bit of a hack, too... :smile: 
As an exercise, I will try and do that in a cleaner way.

I must also remove a console.log() I forgot.

PhiLhoSoft:
OK, I replaced reduce with observe, looks cleaner, no?
And dragOffset is now local, and attached to the dragInfo object.

briancavalier:
Yes, it looks much nicer.  I do like the `observe` instead of `reduce` since `handleDrag` really is about DOM side-effects rather than actually reducing.

PR #150, opened 2015-07-08
Add browser dist build

briancavalier:
This adds a prepublish script to generate dist/most.js, dist/most.min.js and dist/most.min.js.map (source map) for easy script tag use.  The plan would be then to add these to [jsdelivr](http://www.jsdelivr.com) via its automatic npm-based updating.  So, this should create zero extra work, and rule out any chance of forgetting to create and deploy built versions.


briancavalier:
The intent of this is to make it easy to use/try most in tools such as jsfiddle, jsbin, jsperf, codepen, etc. that still favor script tags.  This isn't really intended as a primary way to consume most.js for an application, although that'll work just fine.

briancavalier:
@nissoh Yeah.  It's kind of weird that `prepublish` also runs on `npm install`.  I guess I'll add the mkdir even though that makes it unix only :/  Might be worth biting the bullet and introducing gulp for this.

briancavalier:
Ok, I switched to web pack instead.  It auto-creates the dist dir.  @nissoh, @sporto, and/or @PhiLhoSoft could you please try it out to see if it works for you?  If so, then the next steps will be to 1) merge this, and 2) send a PR to jsdelivr.  Thanks!

nissoh:
i'm not familiar with jsdelivr.
i think the most common usage client wise would be fetching though bower and eventually will result with dist folder with main.js and main.min.js files.

I will put on a new PR soon

briancavalier:
@nissoh [jsdelivr](http://www.jsdelivr.com) is a popular JS CDN.  The intent of this PR is *only* to provide an easy way to experiment with most.js in tools like jsfiddle, jsperf, etc. that don't support a module system at all.

sporto:
@briancavalier you are ignoring 'dist' in .gitignore, isn't the idea to have that folder in github, so we can refer to dist/most.js using https://rawgit.com/ ?

briancavalier:
@sporto See the PR description.  My plan is not to keep dist in the repo, but rather to use jsdelivr.  It automatically tracks updates to npm.  When we publish a new version, jsdelivr will npm install it, which will have the dist folder (because of the prepublish), and will automatically publish the built files to the jsdelivr CDN.

This is especially nice because it is all automated.  There's no way to forget to build the dist version and commit it (because it never gets committed to begin with).

Sound ok?

sporto:
@briancavalier 	yes, sounds good, but there is nothing to test atm as there is no dif folder and it is not in jsdelivr yet, is there?

briancavalier:
@sporto Yeah, sorry, I haven't done the jsdelivr PR yet.  You can run the prepublish by cloning and just doing `npm install`, which will run prepublish for you locally and build the dist files, if you want to try them in a script tag to verify.

I wanted to get feedback on this approach before I submitted a PR to jsdelivr.  Hopefully I'll do that in the next day or so.

nissoh:
@briancavalier the bower version is very misleading. i've tried to point a few front-end developers/designers. they got very confused on how to import most to their workflow.

My pointers would be the to either create most-bower(will require some maintaining) or perhaps give some guidance on how to incorporate this using cdn(bower should be removed).

briancavalier:
Sorry for the hiatus!

@nissoh I haven't used bower seriously for quite some time.  When I did, however, it was easy to install a modular package (with amd or umd modules) and consume it using a module loader, like requirejs.  I was imagining that's how people would use it.  However, since most.js is authored in commonjs, I can see how that is a bit more difficult, since not many browser module loaders support it.  However, both browserify and webpack do.

If bower were like npm, and had a true artifact repository and prepublish script support, this would be easy.  I'd setup a prepublish to build a web pack bundle that gets published to bower.  But alas ...

So, my latest thinking is that the least of all evils is to keep dist artifacts in the repo, much as it pains me :)  So, instead of jsdelivr, we simply add npm scripts to build dist/most.js and dist/most.min.js in umd format (to support amd, commonjs, and script tag), and *hopefully remember* to run that script before tagging github releases.

I think that solves all the problems without putting any additional burden on consumers:

1. Node users can still npm install because the source gets installed (and the dist folder gets ignored).  Or, alternatively the UMD gets installed, and the src gets ignored.  Either strategy will work just fine in node.
2. Bower users can bower install and get the umd and minified umd versions, which will work in any AMD loader and script tags.
3. Script tag only users can download the dist or dist+min versions or reference them via rawgit.com.
4. Folks can use rawgit.com for non-module-aware tools like jsfiddle, codepen, etc.

Thoughts?

briancavalier:
Ok, I implemented the plan above: npm install hasn't changed, it installs the unbuilt sources. Built artifacts will be in the repo, in the dist folder, and bower install will include dist/*.js and not much else.  Script tag users can point at rawgit.com for exploration and either install from bower or download from the dist folder.

Seems like the least of all evils and the easiest to consume for everyone.

PR #158, opened 2015-09-01
Multicast + Reduce

divmain:
Hi - I'm very much enjoying `most` but have run into a problem when using the new `multicast()` in conjunction with the same stream reduced twice.  I may be able to work around my issue, but thought it worth offering a failing test case.  (CI should fail)

I'm new to the codebase but will poke around to see if I can contribute a fix, unless you beat me to it.  Thanks!

### DIFF
--- a/test/combinator/multicast-test.js
+++ b/test/combinator/multicast-test.js
@@ -2,6 +2,7 @@ require('buster').spec.expose();
 var expect = require('buster').expect;
 
 var streamOf = require('../../lib/source/core').of;
+var fromArray = require('../../lib/source/fromArray').fromArray;
 var map = require('../../lib/combinator/transform').map;
 var observe = require('../../lib/combinator/observe').observe;
 var multicast = require('../../lib/combinator/multicast').multicast;
@@ -47,5 +48,24 @@ describe('multicast', function() {
 			expect(observer1Spy).toHaveBeenCalledOnce();
 			expect(observer2Spy).toHaveBeenCalledOnce();
 		});
+	});
+
+	it('when multicasted and reduced twice, source streams are only evaluated once', function() {
+		var sourceSpy = this.spy();
+		var sumSpy = this.spy(function(sum, val) { return sum + val; });
+		var prodSpy = this.spy(function(prod, val) { return prod * val; });
+
+		var s = fromArray([1, 2, 3]).tap(sourceSpy);
+		var multicasted = multicast(s);
+
+		return multicasted.reduce(sumSpy, 0)
+			.then(function (sum) {
+				return multicasted.reduce(prodSpy, 1);
+			})
+			.then(function (prod) {
+				expect(sourceSpy).toHaveBeenCalledThrice();
+				expect(sumSpy).toHaveBeenCalledThrice();
+				expect(prodSpy).toHaveBeenCalledThrice();
+			});
 	});
 });


briancavalier:
Hey @divmain!  Yeah, this is kind of a weird side-effect of most not having gone all-in on being a reactive event stream library.  Although most.js was conceived as a reactive event stream library, I (unfortunately, imho) included a few APIs, such as `from` (and it's related `fromArray`), that allow creating things that are more like collections or replayable data streams.

In a true event stream, time is significant: an event occurrence (value) cannot be known until its time arrives, and once an its time has passed, the occurrence will never be seen again.  That kinda falls apart when creating a replayable stream from an array, where time is basically irrelevant: all the values are available all the time, and it's not really modeling *events* at all.

For example, consider true events, such as mouse clicks, or waking up each morning, or landing on the moon.  Barring any sort of recording/playback machinery, each event can only be observed once it arrives, has effectively no duration, and then is basically "gone" forever.

From that perspective, `fromArray()` + `multicast()` is, in fact, behaving *more* like a real event stream! :)

All of that said, I'm not really sure what the right answer is moving forward.  The only *clear* answer in my mind, is to decide that most.js is for reactive, discrete event streams only, and shouldn't deal with things that aren't actually events.

Any thoughts on all of that rambling?

shader:
I'd actually be in favor of the change if we could have more and better tools for generating and consuming events. The most-bus / publisher method of stream generation feels like a similar kludge that I'd rather move away from if I could figure out how.

Up to this point I've mostly used 'from' as an easy way to create test cases for things depending on streams. I could probably switch to using a combination of iterate and map to accomplish the same thing though. Perhaps `from` could be thought of as `most.iterate(i => i+1,0).map(i => a[i])`, and if that's not 'reactive' enough you could use `most.periodic(0,1).scan((sum, x) => sum+x, 0).map(i => a[i])`. Now they're technically separate events -- they're just as close to immediately after each other as possible without being synchronous. Given that construction any can be converted into a stream, so my remaining comments are probably unnecessary.

Another thing I've been using `from` for is flattening a stream mapped through a function that returns multiple values as an array. The easiest way to do that seems to be `.chain(a => most.from(a))` Maybe some alternate method would fit the FRP style better?

I've also used it to construct a stream processing function suitable for use with `chain` that maps an array of functions over the current stream value.

divmain:
As mentioned, I'm new to `most` so may be missing something.  However, when I modify the test to the following, the test still fails.

```
		var sourceSpy = this.spy();
		var sumSpy = this.spy(function(sum, val) { return sum + val; });
		var prodSpy = this.spy(function(prod, val) { return prod * val; });

		var s = unfold(function (arr) {
			if (!arr.length) return { done: true };
			return { value: arr[0], seed: arr.slice(1) };
		}, [1, 2, 3]).tap(sourceSpy);
		var multicasted = multicast(s);

		return multicasted.reduce(sumSpy, 0)
			.then(function (sum) {
				return multicasted.reduce(prodSpy, 1);
			})
			.then(function (prod) {
				expect(sourceSpy).toHaveBeenCalledThrice();
				expect(sumSpy).toHaveBeenCalledThrice();
				expect(prodSpy).toHaveBeenCalledThrice();
			});

```

That leads me to believe this issue is unrelated to `most.from`.

briancavalier:
@divmain Actually your example reinforces the point :)  I'll try to explain.  

Another way to think about it is that an event stream is like a timeline that is associated with some clock, for example, the clock on the wall, or the clock in your laptop.  For illustration, a way to visualize that is as a list of (time, value) pairs along with some notion of "now".  Reduce, in effect, starts reducing the stream at events whose time is "now" or later.  Let's say we have these events (i.e. time-value pairs):

`[(1, a), (2, b), (3, c)]`

If "now" is time = 1, then reducing the stream will reduce 3 events end an infinitesimal amount of time after time = 3, iow "now" > 3.  Running another reduce when "now" is at time > 3 will see no events, since there are no events with time > 3.

In your example:

```js
multicasted.reduce(sumSpy, 0)
            .then(function (sum) {
                return multicasted.reduce(prodSpy, 1);
            })
```

The promise returned by the first reduce will fulfill at a time after the last event in `multicasted`.  So, starting another reduce of `multicasted` after that will see no events.  Make sense?

That's why something like `from` is so weird: it's not clear what the associated time of values in an array is, or how they related to "now".

shader:
@divmain I think perhaps you have incorrect assumptions as to the purpose of multicast?

It seems that you are hoping that multicast will 'clone' the streams, so that it can be used in multiple places at separate times, with each use receiving the same events.

From what I understand however, `multicast` is mostly for creating 'checkpoints' in the stream, so that all of the processing done before the call to `multicast` only occurs once. It does nothing to preserve the contents of the stream so that it can be used twice.

In this case you are taking a single stream, `multicasted` and reducing it once. It no longer contains any elements. You then proceed to attempt a second reduction, but it no longer contains anything. Thus I fully expect `sourceSpy` and `sumSpy` to have been called 3 times each, yet `prodSpy` not at all.

This comes back to the issue with non-event based streams. Because all of the events are already queued in the stream before any of the reductions, the first one immediately consumes all of them. If you could add both consumers to the stream before draining it, you should get the results you expect. Unfortunately, the only way to do that is to add the consumers before adding any events, because the mere act of adding an observer, reduce, etc. causes the stream to drain.

Most.js streams are lazy, and don't really exist until observed, reduced, etc. That can make some things about them confusing. One thing you could try would be to split the stream before observing it by using `scan` instead of `reduce`. E.g.:
```javascript
var sums = multicasted.scan(sumSpy, 0)
var prods = multicasted.scan(prodSpy, 1)
```
Then you should be able to use `drain` or `reduce((acc, x)=>x, null)` separately on each of the streams. @briancavalier might be able confirm that. I really only have a high-level and rudimentary understanding of most of most.

divmain:
@briancavalier - thanks, that was helpful.  I had completely missed that the second reduce would be invoked after all events had been consumed - totally makes sense.

> From what I understand however, multicast is mostly for creating 'checkpoints' in the stream, so that all of the processing done before the call to multicast only occurs once. It does nothing to preserve the contents of the stream so that it can be used twice.

@shader - you're likely right that my expectations don't match up with the intent here.  However, 

> In this case you are taking a single stream, multicasted and reducing it once. It no longer contains any elements. You then proceed to attempt a second reduction, but it no longer contains anything. Thus I fully expect sourceSpy and sumSpy to have been called 3 times each, yet prodSpy not at all.

This is not the observed behavior either.  `sourceSpy` is called 6 times, and `sumSpy` and `prodSpy` are invoked three times each.  If it were 3, 3, and 0 respectively, I probably wouldn't have been quite so surprised.

The following does result in three invocations for each of the spies:
```
return Promise.all([
	multicasted.reduce(sumSpy, 0),
	multicasted.reduce(prodSpy, 1)
]).then(function (results) {
	expect(sourceSpy).toHaveBeenCalledThrice();
	expect(sumSpy).toHaveBeenCalledThrice();
	expect(prodSpy).toHaveBeenCalledThrice();
});
```

In my particular case (not the test in this PR), the second reduce actually depends on the result of the first.  It should be a straightforward refactor to make it a single reduce.  However, I suspect that readability will suffer.

I suppose what I'm looking for is something similar to `multicast`, but that would queue any events after the method was invoked and make those same events available in order for a second "slower" consumer.

shader:
@divmain Ahh. That's probably because of the ephemeral nature of most.js lazy streams. I told you they confused me sometimes.

Since your first reduce finishes, the stream is considered complete and ceases to exist. Then you add a new observer to the original stream, and it comes back into existence as if the other had never happened. Thus the apparent lack of checkpointing by the multicast. That's a possibility anyway, I know I've run into similar issues before. One solution is to make sure you always have an observer somewhere upstream so that it doesn't pop in and out of existence on you.

divmain:
It seems like the following should work:

```
function fork (strm) {
  return create(function (add, end, error) {
    strm
      .multicast()
      .observe(function (el) { add(el); })
      .then(end)
      .catch(error);
  });
}

// ...

var sourceSpy = this.spy();
var sumSpy = this.spy(function(sum, val) { return sum + val; });
var prodSpy = this.spy(function(prod, val) { return prod * val; });

var s = s.from([1, 2, 3]).tap(sourceSpy);
var a = fork(s);
var b = fork(s);

return a.reduce(sumSpy, 0).then(function () {
    return b.reduce(prodSpy, 1);
}).then(function () {
    expect(sourceSpy).toHaveBeenCalledThrice();
    expect(sumSpy).toHaveBeenCalledThrice();
    expect(prodSpy).toHaveBeenCalledThrice();
});
```

There is a significant downside, in that all values in `s` will be consumed even if `a` and `b`'s consumers don't need all the values, but that's not an issue in my case.

@briancavalier, am I still missing something in expecting the above to work?  Really appreciate your helpful responses (and @shader's)

divmain:
It seems like the following should work:

```
function fork (strm) {
  return create(function (add, end, error) {
    strm
      .multicast()
      .observe(add)
      .then(end)
      .catch(error);
  });
}

// ...

var sourceSpy = this.spy();
var sumSpy = this.spy(function(sum, val) { return sum + val; });
var prodSpy = this.spy(function(prod, val) { return prod * val; });

var s = fromArray([1, 2, 3]).multicast().tap(sourceSpy);
var sA = fork(s);
var sB = fork(s);

return sA.reduce(sumSpy, 0).then(function () {
  return sB.reduce(prodSpy, 1);
}).then(function () {
  expect(sourceSpy).toHaveBeenCalledThrice();
  expect(sumSpy).toHaveBeenCalledThrice();
  expect(prodSpy).toHaveBeenCalledThrice();
});
```

There is a significant downside, in that all values in `s` will be consumed even if `a` and `b`'s consumers don't need all the values, but that's not an issue in my case.

@briancavalier, am I still missing something in expecting the above to work?  Really appreciate your helpful responses (and @shader's).

briancavalier:
@divmain I apologize for letting this sit!  It looks like you've gone a different route in interlock, so is it ok to close this?

divmain:
Thanks for following up - yes, it's okay to close.  I may circle back on this at a later time with a new project.

PR #163, opened 2015-10-18
Make scan preserve end value

### LINKED ISSUE #162: scan should not affect terminal event
Given:

```js
most.create((add, end) => {
  add(2)
  add(4)
  end(6)
})
.scan(
  (acc, x) => {
    return String(acc) + String(x)
  },
  9
)
.observe(console.log)
.then((terminal) => {
  console.log('\nterminal value:')
  console.log(terminal)
})
```

We get:

```
9
92
924

terminal value:
924
```

- Why does `9` appear on its own once?
- Why does scan never see `6`?
- Why does `then` never get `6`?

briancavalier:
Scan was clobbering the stream end value by substituting the last value in the stream for the end value.  This makes scan propagate the end value.  As a bonus, it also simplifies the code slightly.

Fix #162

### DIFF
--- a/lib/combinator/accumulate.js
+++ b/lib/combinator/accumulate.js
@@ -51,10 +51,7 @@ ScanSink.prototype.event = function(t, x) {
 };
 
 ScanSink.prototype.error = Pipe.prototype.error;
-
-ScanSink.prototype.end = function(t) {
-	this.sink.end(t, this.value);
-};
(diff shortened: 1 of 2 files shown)


briancavalier:
cc @jasonkuhrt

jasonkuhrt:
Thanks!

PR #167, opened 2015-11-18
added support for arrayLike items

davidchase:
now we can do `most.from(arguments)` or `most.from(nodeList)`, etc

### DIFF
--- a/test/source/from-test.js
+++ b/test/source/from-test.js
@@ -4,10 +4,6 @@ var expect = require('buster').expect;
 var from = require('../../lib/source/from').from;
 var observe = require('../../lib/combinator/observe').observe;
 
-var sentinel = { value: 'sentinel' };
-var other = { value: 'other' };
-
-
 describe('from', function() {
 
 	it('should support array-like items', function() {
--- a/test/source/from-test.js
+++ b/test/source/from-test.js
@@ -1,0 +1,24 @@
+require('buster').spec.expose();
+var expect = require('buster').expect;
+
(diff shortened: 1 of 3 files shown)


test/source/from-test.js
```diff
@@ -0,0 +1,24 @@
...
+
+describe('from', function() {
+
+	it('should support array-like items', function() {
+		function observeArguments(){
+			var result = [];
+			return observe(function(x) {
+				result.push(x);
+			}, from(arguments)).then(function() {
+				expect(result).toEqual(result);
```
briancavalier: This may not be testing what you think it is :)  I think it will always pass even if `from` is broken.  It's mutating `result`, and ends up comparing `result` with itself.  It should probably compare `result` with a known expected (and immutable :) ) array.

davidchase: yeah wow must have been drinking on that one... :beer: sorry will fix

test/source/from-test.js
```diff
@@ -0,0 +1,25 @@
+require('buster').spec.expose();
+var expect = require('buster').expect;
+
+var from = require('../../lib/source/from').from;
+var observe = require('../../lib/combinator/observe').observe;
+
+var sentinel = { value: 'sentinel' };
+var other = { value: 'other' };
```
briancavalier: These two can be removed.  Then, I think this is good to merge!

davidchase:
@briancavalier let me know if anything is need for this one as well

briancavalier:
:+1: Thanks!

PR #168, opened 2015-11-22
remove unnecessary Function#call in ThrottleSink#end

campersau:
Previously of c36db75fb8ddca310014c7e353e3970ba5f720d8 it was also `ThrottleSink.prototype.end   = Sink.prototype.end;`. Was there a reason why this has been changed?


briancavalier:
Thanks @campersau.  I can't see any good reason for the change--seems like it was just a mistake.  Appreciate the fix :)

PR #170, opened 2015-11-25
doc updates

davidchase:
Removed `lift` from TOC and added `flatMapEnd`

@briancavalier I don't know if `flatMapEnd` ascii diagram or description is ideal but I just wanted to get the ball rolling.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -655,8 +655,8 @@ Replace the end signal with a new stream returned by f. Note that f *must* retur
 
 ```
 stream:               -a-b-c-d-e-f->
-stream.take(4):       -a-b-c-d|
-f(end): 		              -1-2-3-4-5->
+stream.take(4):       -a-b-c-d|end
+f(end): 		               1-2-3-4-5->
 stream.flatMapEnd(f): -a-b-c-d-1-2-3-4-5->
 ```
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -27,6 +27,7 @@ most.js API
 	* [constant](#constant)
 	* [scan](#scan)
 	* [flatMap](#flatmap), alias [chain](#flatmap)
+	* [flatMapEnd](#flatmapEnd)
 	* [concatMap](#concatmap)
 	* [ap](#ap)
 	* [timestamp](#timestamp)
@@ -55,7 +56,6 @@ most.js API
 1. Combining streams
 	* [merge](#merge)
 	* [combine](#combine)
-	* [most.lift](#mostlift)
 	* [sample](#sample)
 	* [sampleWith](#samplewith)
 	* [zip](#zip)
@@ -642,6 +642,33 @@ most.from([1, 2])
 		return most.periodic(x * 1000).take(5).constant(x);
 	})
 	.observe(console.log.bind(console));
+```
+
+### flatMapEnd
+
+####`stream.flatMapEnd(f) -> Stream`
+####`most.flatMapEnd(f, stream) -> Stream`
+
+Map the end event to a new stream, and begin emitting its values.  Note that `f` *must* return a stream.
+
+`function f(x) -> Stream`
+
+```
+stream:            -a-b-c-d-e-f->
+stream.take(4):         -a-b-c-d|
+f(end): 				-1-2-3-4-5->
+stream.flatMapEnd(f): -a-b-c-d-1-2-3-4-5>
+```
+
+
+```js
+most.periodic(1000, 'x')
+  .take(4)
+  .flatMapEnd(function(x){
+    return most.just('hello captain!')
+  })
+  .observe(console.log.bind(console));
+  // Logs: x 4 times... ends and then logs 'hello captain!'
 ```
 
 ### concatMap
--- a/docs/api.md
+++ b/docs/api.md
@@ -27,7 +27,7 @@ most.js API
 	* [constant](#constant)
 	* [scan](#scan)
 	* [flatMap](#flatmap), alias [chain](#flatmap)
-	* [flatMapEnd](#flatmapEnd)
+	* [flatMapEnd](#flatmapend)
 	* [concatMap](#concatmap)
 	* [ap](#ap)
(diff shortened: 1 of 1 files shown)


briancavalier:
Awesome.  Very much appreciated, @davidchase!  Taking a look now.

docs/api.md
```diff
@@ -27,6 +27,7 @@ most.js API
 	* [constant](#constant)
 	* [scan](#scan)
 	* [flatMap](#flatmap), alias [chain](#flatmap)
+	* [flatMapEnd](#flatmapEnd)
```
briancavalier: `#flatmapEnd` -> `#flatmapend`

docs/api.md
```diff
@@ -644,6 +644,33 @@ most.from([1, 2])
...
+Map the end event to a new stream, and begin emitting its values.  Note that `f` *must* return a stream.
+
+`function f(x) -> Stream`
+
+```
+stream:            -a-b-c-d-e-f->
+stream.take(4):         -a-b-c-d|
+f(end): 				-1-2-3-4-5->
+stream.flatMapEnd(f): -a-b-c-d-1-2-3-4-5>
+```
```
briancavalier: Hmmm, after looking at the rendered markdown, I wonder if it's more clear like this:

```js
stream:               -a-b-c-d-e-f->
stream.take(4):       -a-b-c-d|
f(end):                       -1-2-3-4-5->
stream.flatMapEnd(f): -a-b-c-d-1-2-3-4-5->
```

What do you think?  It'd be nice to find a way to represent the end signal value in these diagrams.  A while back, someone had proposed something like the following:

```js
stream:               -a-b-c-d-e-f->
stream.take(4):       -a-b-c-d|x
f(x):                          1-2-3-4-5->
stream.flatMapEnd(f): -a-b-c-d-1-2-3-4-5->
```

I don't think we need to tackle that in this PR (since it has far-reaching documentation implications!), but just wanted to see what you think.

docs/api.md
```diff
@@ -644,6 +644,33 @@ most.from([1, 2])
 	.observe(console.log.bind(console));
 ```
 
+### flatMapEnd
+
+####`stream.flatMapEnd(f) -> Stream`
+####`most.flatMapEnd(f, stream) -> Stream`
+
+Map the end event to a new stream, and begin emitting its values.  Note that `f` *must* return a stream.
```
briancavalier: I've been trying to avoid imperative terminology like "begin emitting", in favor of declarative terminology.  Maybe we can describe this as a form of concatenation.  IOW, conceptually, `flatMapEnd` concatenates `stream` and `f(end)` (even though the value of `end` can't be known until the future when it occurs!).  Thoughts?

docs/api.md
```diff
@@ -644,6 +644,33 @@ most.from([1, 2])
...
+f(end): 				-1-2-3-4-5->
+stream.flatMapEnd(f): -a-b-c-d-1-2-3-4-5>
+```
+
+
+```js
+most.periodic(1000, 'x')
+  .take(4)
+  .flatMapEnd(function(x){
+    return most.just('hello captain!')
```
briancavalier: I think it might be more helpful to show `f` returning a stream with many events--something other than `just`--maybe another `periodic` or something similar.

docs/api.md
davidchase: yes i like the 1st diagram with the `f(end)` i dont know if the `x` after the end is needed?

docs/api.md
davidchase: i agree... i originally had it return another `periodic` just wanted to use something different but for illustration purposes a stream with many events is key.. and will go well with the ascii diagram

docs/api.md
davidchase: What would be another term other than `emitting` just wondering and concat is not a bad idea but its like you said the `end` is unknown so do you concat with a future value?

briancavalier: "Extend" is another pretty good word, but not sure how to incorporate it.  Any ideas?

Here's another angle with "replace".  Maybe with the ascii diagram, this is good enough:

"Replace the end signal with a new stream returned by `f`.  Note that `f` *must* return a stream."

docs/api.md
briancavalier: End-of-stream actually carries a value, which is why I tried showing the `x`.  It's not really well-documented anywhere else, so yeah I agree, we should probably go with the first diagram for now.

docs/api.md
briancavalier: > and will go well with the ascii diagram

:+1: Sounds good

docs/api.md
davidchase: Maybe _extending_ the tail end of one stream with the start of a new one? :stuck_out_tongue_winking_eye: 
> "Replace the end signal with a new stream returned by f. Note that f must return a stream."

nicely put... i like it

docs/api.md
davidchase: Ahh yes that makes and i know `X` is symbolized as an error but even the little `x` makes me think of an error

briancavalier: Yeah, good point.  Maybe we should just pick a different letter?

briancavalier:
This seems really close @davidchase.  Let me know if you need anything else from me.

davidchase:
yup @briancavalier sorry, will update shorty :)

davidchase:
Back to the conversation about an alternative to "x" for end of stream how about the Greek letter Ω ?

Or do you not prefer to introduce Greek lettering in the docs?

briancavalier:
I don't mind the Ω, but I could see it causing some double-takes :)  Maybe we just go with your original idea of `end`, iow: `-a-b-c-d|end` and `f(end)`.  Then we just rely on the prose to describe it?

davidchase:
Yeah that works I was throwing stuff out to see what sticks.

Should I update the PR?

briancavalier:
Yep, thanks!

davidchase:
@briancavalier all good now :smiley_cat:

briancavalier:
:tada:

PR #174, opened 2015-12-14
Add mergeConcurrently to public API

briancavalier:
It's used to implement join, flatMap, and concatMap, so is a generally useful building block.  I don't see any reason not to make it available.  We can bikeshed the name :)

Still needs:

- [x] docs
- [x] unit tests

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -62,6 +62,7 @@ most.js API
 1. Combining higher order streams
 	* [switch](#switch)
 	* [join](#join)
+	* [mergeConcurrently](#mergeConcurrently)
 1. Awaiting promises
 	* [await](#await)
 1. Rate limiting streams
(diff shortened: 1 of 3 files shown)


briancavalier:
cc @axefrog

axefrog:
Cool. Might be worth mentioning something regarding the potential for an infinite backlog if backpressure is not managed.

briancavalier:
> Might be worth mentioning something regarding the potential for an infinite backlog

Yeah, for sure.  I was thinking we could handle that via prose in the full API docs.  Sound ok?

> capping the queue size, deciding whether the tail or the head should be discarded at capacity, and possible even LIFO/FIFO options

Excellent thoughts.  I've done things like this in the past by allowing the caller to provide functions rather than "options".  It might have to be a pair of functions in this case: a function that describes the behavior when a new incoming stream arrives, and a function that describes the behavior for when an currently-active stream ends.  Any thoughts on that?

axefrog:
I'm all for a functional approach, if it makes sense. Whatever you think provides the best opportunities for composability.

briancavalier:
@axefrog thinking about this a bit more, rather than holding up this PR, I'm thinking the right thing is to finish the docs and tests, and merge this as is.  If a more general/flexible operation becomes obvious, we can always add it and likely change the implementation of `mergeConcurrently` to use it.  What do you think?

axefrog:
Hey, my bad I completely forgot to test this out. Your assessment seems reasonable to me.

PR #176, opened 2015-12-17
Improve scan initial value handling.

### LINKED ISSUE #175: scan fails to emit seed value immediately
As I understand it, the default/seed value provided to `scan` should be emitted immediately upon consuming the stream. Instead, it appears to be waiting for the first received item before emitting anything, which means nothing is emitted to begin with, and when an item of data is eventually received, the seed value is then emitted, immediately followed by the first transformed data item.

briancavalier:
This also vastly speeds up startWith and concat by relying on flatMapEnd instead of concatMap.

Fix #175

### DIFF
--- a/lib/combinator/accumulate.js
+++ b/lib/combinator/accumulate.js
@@ -5,6 +5,7 @@
 var Stream = require('../Stream');
 var Pipe = require('../sink/Pipe');
 var runSource = require('../runSource');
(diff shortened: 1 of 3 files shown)


briancavalier:
@axefrog Could you try this branch to see if it does what you expect?  Thanks!

PR #186, opened 2016-01-22
Add TypeScript Definition File

TylorS:
Adds TypeScript Definition file and a 'typings' key in the package.json to point to that file for TypeScript users to benefit from its tooling.


TylorS:
Don't merge this yet, TypeScript is not yet able to detect the definitions as is. Not sure why, but I'm looking into it

briancavalier:
Ok, np.  In the meantime, I'll review the type definitions themselves.

TylorS:
So I have some information that is both good and bad. Good, because I figured out how to properly write the TypeScript definition files to where they work. Bad, because you absolutely must have the `.d.ts` file next to the imported file. Even with `typings` pointing to `type-definitions/most.d.ts` TypeScript's tooling doesn't seem to care unless it is next to `most.js`.

TylorS:
After writing that, I'm curious now, if it has more to do with the TypeScript support in Atom, than TypeScript itself. Gonna download VS Code, and see if there is any change.

TylorS:
Yeah, so the tooling doesn't seem to like the same things :-1: 
What works in Atom with `Atom-TypeScript` doesn't working in VS Code

briancavalier:
Ugh. Is there any spot that works for all those tools? For example if we put it beside most.js does that work in a broader set of tools?

TylorS:
I'm really not sure just yet

TylorS:
Okay, I have been able to get it working across editors *with* using `typings` :smile: 

It's not as pretty as I expected. It's necessary to export individual functions in order to do `import {fn} from 'most'` and it's required to also redefine the same functions inside of a namespace to export is all as default. Otherwise `most.fn` isn't recognized. This seems very strange. This *does* work, but I feel like there might be a better way.

briancavalier:
Maybe looking through examples at [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) will shed some light.

TylorS:
That is slightly different. DefinitelyTyped repos are used via references. At the top of the file `/// <reference path="../path/to/most.d.ts"/>. It's not automatic like the solution I was going for, and would probably be best published to DefinitelyTyped to be handled by something like tsd. I prefer the former, but if you think it best to go for the latter I could give it a shot.

briancavalier:
@TylorS Sorry, I should have been more clear.  I was hoping some of the examples there might shed light on how to structure the *contents* of the ts file such that it doesn't have to declare everything twice.

TylorS:
Not that I've seen just yet. I'll keep trying.

TylorS:
Once I figure it out, I won't have this problem again lol

TylorS:
Okay, so it can be declared once if we make it 'best' practice to do
```typescript
import * as most from 'most'
```
instead of
```typescript
import most from 'most'
```

That would only be 'required' for TypeScript users.
Either that or to strictly import the functions that are required (better IMO)
```typescript
import {just, combine} from 'most'
```

briancavalier:
Ahhh, I understand now!  I'm totally fine with that because most *really doesn't* have a default export.  If we get feedback from the TS community that we're doing it wrong, then we can ask for help :)  Sound ok?

Thanks for the continued research!

TylorS:
That seems acceptable to me. I'll push the changes shortly.

TylorS:
I can confirm that this works with Atom, VS Code, and WebStorm 11.

type-definitions/most.d.ts
```diff
@@ -0,0 +1,144 @@
+declare type SeedValue<S, V> = { seed: S, value: V };
+declare type TimeValue<V>    = { time: number, value: V };
+
+declare interface Promise<A> {}
```
briancavalier: Is it possible to reference [the type definition for es6-promise](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/es6-promise/es6-promise.d.ts) without actually listing es6-promise as a dep and without committing its type def file to the most repo? It'd be nice so that callers get the benefit of most.js having well-typed promise values.

TylorS: I'll look into it soon :+1:

type-definitions/most.d.ts
```diff
@@ -0,0 +1,144 @@
...
+  sample<B>(f: (...as: any[]) => B, ...ss: Array<Stream<any>>): Stream<B>;
+  sampleWith<A>(sampler: Stream<any>): Stream<A>;
+
+  zip<A>(f: (a: A, ...as: any[]) => A, ...ss: Array<Stream<any>>): Stream<A>;
+
+  recoverWith<B>(p: (a: B) => A): Stream<A>;
+  multicast<A>(): Stream<A>;
+}
+
+export function create<A>(f: (add: Function, end: Function, error: Function) => void): Stream<A>;
```
briancavalier: I think we can make `create` a bit tighter.  My TS syntax is probably wrong, but hopefully gives the idea:

```ts
export function create<A>(f: (add: (a:A) => any, end: (x:any) => any, error: (e:Error) => any) => (?() => ?Promise<any>)): Stream<A>;
```

The most important thing, IMHO, is tying `add`'s parameter type, to the type parameter of the returned `Stream<A>`.  Also, the function passed to `create` can optionally return a cleanup function, which can optionally return a Promise to represent an async cleanup process.

We might be able to try to incorporate the stream end value as one of the `Stream` type parameters, e.g. `Stream<A, Z>`, but I don't think we need to go there for the first version of this :)

Did that make sense?

type-definitions/most.d.ts
```diff
@@ -0,0 +1,144 @@
...
+export function skipUntil<A>(signal: Stream<any>, s: Stream<A>): Stream<A>;
+export function during<A>(timeWindow: Stream<Stream<any>>, s: Stream<A>): Stream<A>;
+export function throttle<A>(period: number, s: Stream<A>): Stream<A>;
+export function debounce<A>(period: number, s: Stream<A>): Stream<A>;
+
+export function timestamp<A>(s: Stream<A>): Stream<TimeValue<A>>;
+export function delay<A>(dt: number, s: Stream<A>): Stream<A>;
+
+export function fromPromise<A>(p: Promise<A>): Stream<A>;
+export function awaitPromises<A>(s: Stream<Promise<A>>): Stream<A>;
```
briancavalier: Is it possible to declare that there's also an export named `await`?  Or will TS complain about that name?

briancavalier: Hrm, actually, there's currently no `most.awaitPromises`.  We should probably add it, tho.  I'll do that in the morning.

briancavalier:
I think we're getting close!  Just a few questions, and probably need to add `most.awaitPromises`.

type-definitions/most.d.ts
TylorS: Yes, that all made sense :+1:

type-definitions/most.d.ts
TylorS: I really just stole is from your flowtype declarations:
https://gist.github.com/briancavalier/e288e28e2bc2d2c84d91#file-most-js-L62

Figured you had a reason for it

briancavalier: Heh, oops! Sorry about that :) Is it possible to declare the name `await` in TS? I'll check flow also to see if it complains

TylorS:
I made the changes. TS is totally okay with `await()`, I went with a very similar form of `create()` that you mentioned above. I also inlined the Promise interface from es6-promise. I'm not sure if that's what you were hoping for though.

briancavalier:
> I also inlined the Promise interface from es6-promise

I have no idea if there's a better way.  They're surprisingly minimal, though, so it seems fine.  One question though: will TS consider the `Promise` type declared in `most.d.ts` to be a *different type* than a `Promise` type declared somewhere else?  For example, say a codebase uses both most and es6-promise.  Will TS complain if someone tries to pass a "most Promise" where an "es6-promise Promise" is expected?

If that's not an issue, then I think this is good to merge.

briancavalier:
Another thought: we could PR these type definitions to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) and then we'd be able to `/// <reference path="{path/to/es6-promise.d.ts}`.

Hmmm, what do you think of doing that instead of keeping them here in the repo?  Seems a lot of libs take that approach.

TylorS:
I think that it should be okay as long as the promise implement the ES6 promise interface (most of them at this point?) There is a good chance that it probably would always use a the promise defined locally. 

Regarding DefinitelyTyped, I sort of dislike it in some ways (though I'm not totally opposed). It's just way easier to get started using them this way. e..g. a user just has to `npm install` and they're good. With DefinitelyTyped you have to install *tsd* and *tsd install* the definitions and add the references etc. Also with DefinitelyTyped it's way more 'convenient' to let them fall behind. 

However, in favor of DefinitelyTyped is we can consolidate all of the proposed libraries into one file. 
```typescript
declare module "@most/core" {
  // defs here
}

declare module "@most/hold" {
  function hold(...)....
}

declare module "@most/dom-event" {
 ...
}
```

Which allows for one place to keep everything up to date, there would be no need to update @most/whatever's package just to point to the definitions of a new version of @most/whatever if there is some kind of shared interface.

TylorS:
Thinking more about it, DefinitelyTyped might be more useful in that way. Besides, if a user is creating much of anything they probably have to use *something* from there anyways.

briancavalier:
> if a user is creating much of anything they probably have to use something from there anyways

Good point!

> Also with DefinitelyTyped it's way more 'convenient' to let them fall behind

Yeah, that's a valid concern.  It may have an upside, which is that we may get more help from the community in keeping them up to date.

Another related snag is versioning.  If the TS defs are in the most.js repo, then they get versioned directly with the source code.  It's not clear to me how to version the TS defs if they live on DT.  It looks like some libs resort to putting multiple copies of their TS defs on DT, including a version number in the filename.  I found a few discussions, like [this one](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/2972), but haven't found anything definitive yet.

> we can consolidate all of the proposed libraries into one file

That is convenient, but would it make things easier more difficult if the libs change at different rates?  I'm not entirely sure yet.

TylorS:
> That is convenient, but would it make things easier more difficult if the libs change at different rates? I'm not entirely sure yet.

Yeah, you're right. I hadn't thought about versioning. So far it's been just implicit that it's up-to-date being in the repo.

AlexGalays:
It's also more convenient when a library comes with its typescript definition already. No extra step needed at all for the consumers.

briancavalier:
Thanks @AlexGalays.  I think I'm pretty much sold on keeping them in the most.js repo :)

TylorS:
Hey @briancavalier, it's been a while since I kept up with this. Whenever you have the time, if you could point me in the right directions I'd love to get this finished up :smile:

briancavalier:
@TylorS thanks for pinging.  You bet.  I'll look at it soon!

briancavalier:
This looks good!  Any you can think of that still needs to be done, @TylorS ?

TylorS:
@briancavalier I can't think of any 'better' way to do it without adding dev deps to allow importing es6-promise

AlexGalays:
combineArray seems to be missing. it could be strongly typed for a limited number of streams by using typescript's tuples.

By the way, why not using this style of declaring the typings?  https://github.com/AlexGalays/fluxx/blob/master/typings/fluxx.d.ts

This way, types are imported automatically when importing the main module instead of having to manually link the typings file in tsconfig.

briancavalier:
> combineArray seems to be missing. it could be strongly typed for a limited number of streams by using typescript's tuples.

Ah, nice catch. Thank you, and thanks for the suggestion to use tuples with multiple signatures.  Seems reasonable.  @TylorS any thoughts?

> By the way, why not using this style of declaring the typings?

I'm a typescript novice :)  Can you explain a bit more about which "style" you're referring to?  When I look at that fluxx.d.ts file, it's not clear to me how it "automates" the importing of types.  Is it because it's in a directory named "typings"?

Thanks for the help!

TylorS:
:+1: for noticing I'm missing `combineArray`

AlexGalays:
The typings file is linked in package.json : https://github.com/AlexGalays/fluxx/blob/master/package.json#L5
But I saw you did the same. Does it resolve all the types of `most` without linking that typing file in another project?

typescript has 2 or 3 generations of modules and type declaration formats; I lost myself in it quite a few times. I think the style I use is called "external definition" whereas the one you use is called "ambient definition" and they now encourage the former. I also think/thought only "external definitions" can be resolved as a regular npm module by typescript.

TylorS:
@AlexGalays Yeah, this file can be automatically detected via npm

AlexGalays:
Well then, no reason to change it.

briancavalier:
> I think the style I use is called "external definition" whereas the one you use is called "ambient definition" and they now encourage the former

Fascinating!  Are there specific things in the file in this PR vs your fluxx.d.ts that distinguish the two formats?  IOW, is there a way I could learn to recognize an "external definition" by its contents?

TylorS:
I'd also like to second @briancavalier's question, I still have a lot to learn about TypeScript as well.

AlexGalays:
To be honest I thought their only noticeable difference was their compatibility on npm :)

AlexGalays:
So, what's left to be done before this is merged? :)

briancavalier:
@AlexGalays I think just adding `combineArray`.

TylorS:
@briancavalier Just added combineArray and some others I realized were missing

type-definitions/most.d.ts
```diff
@@ -90,11 +93,14 @@ declare interface DisposeFn {
...
 
-export function create<A>(f: (add: (a:A) => any, end: (x:any) => any, error: (e:Error) => any) => void|DisposeFn): Stream<A>;
+export function create<A>(f: (add: (a: A) => any, end: (x: any) => any, error: (e: Error) => any) => void|DisposeFn): Stream<A>;
 export function just<A>(a: A): Stream<A>;
 export function of<A>(a: A): Stream<A>;
+export function empty(): Stream<any>;
+export function never(): Stream<any>;
 export function from<A>(as: Iterable<A>): Stream<A>;
 export function periodic<A>(period: number, a?: A): Stream<A>;
+export function fromEvent(event: string, target: any, useCapture?: boolean): Stream<Event>;
```
briancavalier: Ah, this one is tricky!  Is it possible to provide two signatures for fromEvent?  i.e. since it handles both EventTarget and EventEmitter, we may be able to gain a small amount of type safety if TS supports polymorphic functions ... something like:

```js
export function fromEvent(event: string, target: EventTarget, useCapture?: boolean): Stream<Event>;
export function fromEvent(event: string, target: EventEmitter): Stream<any>;
```

I can't see a way to avoid the `any` in the EventEmitter version's return type :(

Anyway, if that's not possible, then maybe we can use a union type for `target`, e.g. `EventTarget | EventEmitter`.

If that still won't work because TS doesn't know about EventTarget and EventEmitter, then maybe we can define structural types for them?  i.e. at least define what methods they must have?

type-definitions/most.d.ts
```diff
@@ -44,7 +44,10 @@ export interface Stream<A> {
   concatMap<B>(f: (a: A) => Stream<B>): Stream<B>;
   mergeConcurrently<B>(concurrency: number): Stream<B>;
   merge(...ss: Array<Stream<A>>): Stream<A>;
+  mergeArray(streams: Array<Stream<A>>): Stream<A>;
   combine<B>(f: (a: A, ...args: Array<any>) => B, ...ss: Array<Stream<any>>): Stream<B>;
+  combineArray<B>(f: (...args: Array<any>) => B, streams: Array<Stream<any>>): Stream<B>;
```
briancavalier: Should we define a bunch of polymorphic variants for combine and combineArray, [similar to what Rx 5 has done](https://github.com/ReactiveX/rxjs/blob/master/src/operator/combineLatest.ts#L70)?  It's visually awful, but provides some extra type safety when the argument list (or array) is small.

type-definitions/most.d.ts
TylorS: Hmmm right, TS isn't very good for polymorphic functions :/
It also doesn't have a builtin for EventEmitter so maybe it would be best to define some interfaces?

briancavalier: > maybe it would be best to define some interfaces

Sounds good :+1:  If we do that, and we differentiate the two signatures (by not allowing `useCapture` to be used with EventEmitter, since that's pointless), I think it'll be in a pretty good spot.

type-definitions/most.d.ts
TylorS: Yuck, yeah we probably should, but also for zip?

briancavalier: Good catch about zip.  Yeah, I suppose so.

AlexGalays: +1; It's visually awful but people won't go look at the typings too much I'd think, and it greatly helps to have type safe combinators. Up to 6-10 parameters sounds fine.

type-definitions/most.d.ts
AlexGalays: I think this will work as intended as EventTarget and EventEmitter have very different structural interfaces.

type-definitions/most.d.ts
AlexGalays: merge and mergeArray have a wrong signature too. You can merge Arrays of streams with different types. They work similarly to combine/combineArray at the type level. 
I can help with these if you want.

TylorS: @AlexGalays Help would be wonderful, I'm not any good with types and definitely not TypeScript

AlexGalays: cool; I will try to do that this weekend; should I fork your own fork then? :)

TylorS: Yeah that can work, just PR to my fork and this PR will be updated automatically :+1: 

Thanks!!!

briancavalier: > merge and mergeArray have a wrong signature too

I think there is a question here.  Allowing `merge(Stream<A>, Stream<B>): Stream<A | B>` makes it easy to create heterogeneous streams.  One of the goals of types is to provide guidance and help users avoid mistakes.  While heterogeneous streams are *possible*, I'd like to weigh their pros and cons to understand if they are helpful.

Personally, I tend to always use homogeneous streams, and haven't run into a case where I needed a heterogeneous stream.

What are some *concrete* use cases for heterogeneous streams?

Looking around at some strongly typed haskell implementations:

1. Reactive Banana's [unionWith](http://hackage.haskell.org/package/reactive-banana-1.1.0.0/docs/Reactive-Banana-Combinators.html#v:unionWith) is homogeneous.
2. Netwire's [merge, mergeL, and mergeR](http://hackage.haskell.org/package/netwire-5.0.1/docs/Control-Wire-Event.html#v:merge) are homogeneous.
3. FRP.Reactive's [eitherE](http://hackage.haskell.org/package/reactive-0.11.5/docs/FRP-Reactive.html#v:eitherE) is homogeneous in it's output: it *always* produces something of type `Either a b` (it's a subtle but important distinction that it does not produce something of either type 'a' or type 'b').  However, it certainly has a heterogeneous flavor to it.

Thoughts?

type-definitions/most.d.ts
briancavalier: I should never have conflated the two in the first place, but for now, we're stuck with my unfortunate, dual-purpose `fromEvent` :)

Given that, I do think it's important to provide the extra guidance of only allowing `useCapture` when `target` is an `EventTarget`.  While `fromEvent('blah', anEventEmitter, true)` isn't harmful, it also doesn't do anything useful beyond `fromEvent('blah', anEventEmitter)`.

We should also try to help people discover and use [`@most/dom-event`](https://github.com/mostjs/dom-event) instead of `fromEvent` for DOM events whenever possible.

type-definitions/most.d.ts
AlexGalays: That's a fair point! Forgot how merge worked for a second apparently ;)
In a type safe environment, you would definitely want homogeneous streams for sure, even more so than in plain JS.
Developers can always type their 2 streams as Stream<A | B> if they want to.

jzhang1:
Is there a most.d.ts somewhere?

briancavalier:
@jzhang1 That's the goal of this PR :)  So, when this is merged, it will be available in the repository and npm package.  If you want to try it out before then, you could grab it out of this PR branch.

type-definitions/most.d.ts
```diff
@@ -4,7 +4,7 @@ declare type TimeValue<V>    = { time: number, value: V };
 declare interface Thenable<A> {
     then<U>(onFulfilled?: (value: A) => U | Thenable<U>, onRejected?: (error: any) => U | Thenable<U>): Thenable<U>;
     then<U>(onFulfilled?: (value: A) => U | Thenable<U>, onRejected?: (error: any) => void): Thenable<U>;
-    catch<U>(onAejected?: (error: any) => U | Thenable<U>): Thenable<U>;
+    catch<U>(onRejected?: (error: any) => U | Thenable<U>): Thenable<U>;
```
briancavalier: 😄

type-definitions/most.d.ts
```diff
@@ -57,8 +88,8 @@ export interface Stream<A> {
   startWith(a: A): Stream<A>;
 
   filter(p: (a: A) => boolean): Stream<A>;
-  skipRepeats(): Stream<A>;
-  skipRepeatsWith(eq: (a1: A, a2: A) => boolean): Stream<A>;
+  skipBepeats(): Stream<A>;
+  skipBepeatsWith(eq: (a1: A, a2: A) => boolean): Stream<A>;
```
briancavalier: Bepeats! Skip all the bepeats!

TylorS: Lol find and replace fail :)

TylorS:
It now only skips repeats :+1:

TylorS:
I think I need to add `thru()` to this now as well

fejesbence:
It seems you still have some "bepeats" :)

AlexGalays:
Only a typo fix for this to be merged ? :p

AlexGalays:
@briancavalier  I think this is good to go?

briancavalier:
@AlexGalays Thank you!  I'll make time to look at this later today.

type-definitions/most.d.ts
```diff
@@ -0,0 +1,349 @@
...
+  tap(f: (a: A) => any): Stream<A>;
+  chain<B>(f: (a: A) => Stream<B>): Stream<B>;
+  flatMap<B>(f: (a: A) => Stream<B>): Stream<B>;
+
+  ap<B, C>(fs: Stream<(a: A) => B>): Stream<C>;
+
+  // Note: Without higher-kinded types, the types for these
+  // cannot be written in a completely safe manner; See https://github.com/Microsoft/TypeScript/issues/1290
+  // For better type safety, consider using most.join/switch/switchLatest with thru
+  join(): A;
```
briancavalier: Thanks for including the link to the TS issue!  Let's go with it until there's a better option :)

type-definitions/most.d.ts
```diff
@@ -0,0 +1,349 @@
...
+    c: Stream<C>,
+    d: Stream<D>
+  ): Stream<R>;
+  combine<B, C, D, E, R>(
+    fn: (a: A, b: B, c: C, d: D, e: E) => R,
+    b: Stream<B>,
+    c: Stream<C>,
+    d: Stream<D>,
+    e: Stream<E>
+  ): Stream<R>;
```
briancavalier: I really like the explicitness of enumerating these. :+1:

Do you think we should include a "catch-all" variant at the end, like:

```ts
combine<R>(
  fn: (...args: Array<any>) => R, ...streams: Array<any>
): Stream<R>
```

Or would that cause more problems than it solves?

type-definitions/most.d.ts
```diff
@@ -0,0 +1,349 @@
...
+
+  combineArray<B, R>(
+    fn: (a: A, b: B) => R,
+    array: [Stream<B>]
```
briancavalier: Not a blocker for merging, but it's not clear to me what to name this parameter.  For example, it's obviously an array from the type signature, so the name `array` is redundant.  I was going to suggest the name `streams`, but that's *also* redundant since it's obviously an array of Stream.

Struggling to find a good name ... `inputs`??

type-definitions/most.d.ts
```diff
@@ -0,0 +1,349 @@
...
+  b: Stream<B>,
+  c: Stream<C>,
+  d: Stream<D>,
+  e: Stream<E>
+): Stream<R>;
+
+export function combineArray<A, B, R>(
+  fn: (a: A, b: B) => R,
+  array: [Stream<A>, Stream<B>]
+): Stream<R>;
```
briancavalier: I really like that you've enforced the minimum number of array elements (as well as varargs above) to 2. :+1:

briancavalier:
@AlexGalays @TylorS I really only had one question about adding "catch-all" variants, plus a small comment about naming (not a blocker to merging).  Let me know what you think about the "catch-all", and we should be able to get this merged.

AlexGalays:
@briancavalier 

A catch-all is counter productive and would make all the other definitions redundant. Typescript tries all the overloaded definitions from top to bottom till one work before throwing an error so if you make a type error such as swapping the types of the second & third streams in the `combine` function compared to the input streams then the catch-all be retained and you won't see a compilation error.

About the naming, I think `streams` is indeed a better name (array is more technical and domain agnostic)

Lastly, we can always further improve the definitions later! For instance, adding JSDoc above each definitions would make these appear in Atom's intellisense, which is quite handy (at the cost of doc duplication with your github page).

briancavalier:
@AlexGalays Thanks for the great explanation.  That makes perfect sense.

It's probably rare for someone to try to combine a large number of streams, but it's probably inevitable (if it's possible, someone will do it) :)  I'm fine with crossing that bridge when we come to it.

> I think streams is indeed a better name

Cool, let's go with it.

> we can always further improve the definitions later!

Yes, absolutely.

Let's rename the `array` params to `streams`, and then merge this!

briancavalier:
Merging!  Awesome work @TylorS and @AlexGalays.  Thank you.

AlexGalays:
I think you didn't republish a new version to npm? Also, the type-definitions folder would have to be added to `files` in package.json

briancavalier:
Good thing we didn't publish it then :)  I'll have time later tonight to update `files` and publish to npm.

briancavalier:
it's released! 💥 https://github.com/cujojs/most/releases/tag/0.19.6 💥

PR #187, opened 2016-01-23
Use native promise

briancavalier:
This removes when.js's es6 shim in favor of using native Promise.  Environments that don't have Promise will need to use a polyfill.

There are probably a few spots that will incur a small perf hit since native Promise can be significantly slower.  One of those was in mergeConcurrently, and I was able to refactor it slightly to avoid the
hit in cases where disposal doesn't return a promise.  If it's a problem, apps can use a fast polyfill like creed, when, or bluebird.

I also switched to eslint because jshint complained about `Promise`, and I didn't want to mess with configuring it to recognize `Promise` as a global.  I've come to like eslint anyway :)


README.md
```diff
@@ -69,6 +69,12 @@ To [run the example above](examples/add-inputs) and [others](examples) using [Ra
 
 ## Get it
 
+### Requirements
+   
+   Most requires ES6 `Promise`.  In environments that don't have ES6 `Promise`, use a polyfill such as [creed](https://github.com/briancavalier/creed), [when](https://github.com/cujojs/when/blob/master/docs/es6-promise-shim.md), [bluebird], etc.
```
TylorS: No link for `bluebird`

test/perf/concatMap.js
```diff
@@ -1,11 +1,10 @@
 var Benchmark = require('benchmark');
 var most = require('../../most');
 var rx = require('rx');
-var rxjs = require('@reactivex/rxjs')
+var rxjs = require('@reactivex/rxjs');
```
TylorS: Oops ;)

TylorS:
Looks good! :+1: :+1:

README.md
briancavalier: :facepalm: thanks!

briancavalier:
Added the missing bluebird link.

briancavalier:
A big win on overall size with this change:  it drops 3k off the *min+gzipped* size:

with when.js shim included:
```
-rw-r--r--  1 brian  staff   13361 Jan 22 14:34 most.min.js.gz
```

using native promise (no when.js shim):
```
-rw-r--r--  1 brian  staff   10105 Jan 25 11:45 most.min.js.gz
```

PR #188, opened 2016-01-25
Expose mergeArray and combineArray

briancavalier:
This exposes the existing array versions of merge and combine, since they can be useful at times, and `.apply(undefined, streams)` is ugly.

Docs and unit test update included.  This also adds a perf test for merge.


TylorS:
Nice!

PR #190, opened 2016-01-28
Safely dispose in the face of exceptions

briancavalier:
Disposal was previously safe in the face of rejected promises.  That is, all other resources would be disposed even if one or more of them failed by returning a rejected promise.  Disposal was not safe in the face of exceptions.

There are 2 ways to handle thrown exceptions during disposal:

1. Declare that disposables *must not throw*, and any disposable that does is broken and should be fixed, or
2. Catch exceptions and turn them into rejected promises, and let the current machinery handle them safely.

This PR does 2.  However, I'm not convinced it's the right thing, since it allows callers to be less rigorous.  So, feedback and discussion welcome!

This also adds unit test coverage for disposables.  If we decide to go with option 1, we can remove the one `try/catch` spot in `disposeSafely`, and keep the unit test coverage (except for the tests that specifically deal with thrown exceptions).

### DIFF
--- a/lib/disposable/dispose.js
+++ b/lib/disposable/dispose.js
@@ -27,7 +27,7 @@ exports.promised = promised;
  * @return {*} result of disposable.dispose
  */
 function tryDispose(t, disposable, sink) {
-	var result = disposable.dispose();
+	var result = disposeSafely(disposable);
 	return isPromise(result)
 		? result.catch(function (e) {
 			sink.error(t, e);
@@ -65,11 +65,15 @@ function all(disposables) {
 }
 
 function disposeAll(disposables) {
-	return Promise.all(map(disposeOne, disposables));
+	return Promise.all(map(disposeSafely, disposables));
 }
 
-function disposeOne(disposable) {
-	return disposable.dispose();
+function disposeSafely(disposable) {
+	try {
+		return disposable.dispose();
+	} catch(e) {
+		return Promise.reject(e);
+	}
 }
 
 /**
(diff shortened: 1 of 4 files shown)


TylorS:
I'm having a total brain cramp trying to look at this today. I've had way too much coffee :smile: 

I promise to look back at this later tonight

briancavalier:
Thanks @TylorS :)

I thought of option 3:

Catch exceptions and next-tick rethrow them to crash the program (real crash on node, loud window.error in browsers).  We could even wrap them with an obvious error message pointing to the broken disposable.  That'd make it painfully obvious that code outside of most.js is broken and must be fixed.

It's basically a variant of option 1 that enforces the constraint that disposables must not throw :)

Thoughts?

TylorS:
> wrap them with an obvious error message 

I think anything that can help make errors more obvious/accessible is a win so if I had to pick between 1 and 3 I'd absolutely choose option 3. That's one of the weakest points regarding stream libraries IMO (and others from what I've seen). 

At the same time I think it's really tough to decide whether or not a program should crash because of it. I can already foresee people wanting to handle the errors themselves and general complaining.

briancavalier:
> I can already foresee people wanting to handle the errors themselves and general complaining.

They still could.  We'd just be enforcing that to *safely and correctly* surface an error, disposers must return a rejected promise.

Another thought on option 2 is that by catching exceptions and turning them into rejections (as the PR does now), the error still does propagate and if you don't `.catch()` it (e.g., via the promise returned from `observe`), a good promise implementation will still surface it (even several native impls do that now).

TylorS:
> They still could

Ahh okay, I think I misunderstood then. 

> a good promise implementation will still surface it

Then I think option 2 makes a lot of sense

briancavalier:
> Then I think option 2 makes a lot of sense

Yeah, I'm coming back around to option 2 :)  There is a natural progression from 2 to 3 (since both require adding `try/catch` to the existing machinery) if, in the future, we want to add the crashing feature :)

PR #194, opened 2016-02-05
Deprecate cycle()

briancavalier:
It doesn't seem to be useful.  If there's a need for it, it could be published as a separate package.

We can release 0.18.1 with the deprecation, and remove it in 0.19.0.


TylorS:
I agree with it not being very useful. :+1:  for deprecation

davidchase:
and since you can build it yourself with `continueWith` theres no reason to keep it :100:

PR #195, opened 2016-02-11
Update reqr.es links in example.

jeffutter:
It looks like reqr.es has done away and the project is now hosted at reqres.in . I'm not sure what the api returned before - I'm assuming it is similar.


briancavalier:
Nice catch, @jeffutter.  Thank you!

PR #199, opened 2016-02-25
Fix scheduler task cancelation bug

### LINKED ISSUE #197: sampleWith breaks when value stream is debounced
`sampleWith` stops sampling events in my current project after one of the input streams changes. The stream transformation looks something like this:

```js
const coordinateStream = most.combineArray(
    (gridSize, viewport, center, zoom) => calculateCoordinates({ gridSize, ...viewport, center, zoom })
  , [gridSizeStream, viewportStream, centerStream, zoomStream]);

const brokenStream = most.combine(
    (init, coordinates) => ({ init, coordinates })
  , initStream, coordinateStream)
  .map(transformation1)
  .map(transformation2)
  .sampleWith(most.periodic(1000))
  .skipRepeats();
```

The weird thing is that it works when `centerStream` or `zoomStream` emit events, but it breaks as soon as `viewportStream` does. I’ve tapped into `brokenStream` before and after `sampleWith` to see if it receives new and correct values and it does, but it just stops sampling when `viewportStream` emits a new event. Removing `sampleWith` fixes the problem, but obviously results in a stream that does emit values more than once a second.

The description might be too quite vague, but I’m not sure what other information I can give that is helpful. Any ideas?

briancavalier:
Fix subtle bug in scheduler task cancelation.

Fix #197

### DIFF
--- a/lib/scheduler/Scheduler.js
+++ b/lib/scheduler/Scheduler.js
@@ -66,7 +66,7 @@ Scheduler.prototype.periodic = function(period, task) {
 
 Scheduler.prototype.schedule = function(delay, period, task) {
 	var now = this.now();
-    var st = new ScheduledTask(now + Math.max(0, delay), period, task, this);
+	var st = new ScheduledTask(now + Math.max(0, delay), period, task, this);
 
 	insertByTime(st, this._tasks);
 	this._scheduleNextRun(now);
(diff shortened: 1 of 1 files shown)


unscriptable:
tricky! :)

PR #203, opened 2016-03-04
Fix scheduler cancelAll

briancavalier:
Fix for case that could cause cancelAll to remove the wrong tasks from the scheduler queue.

Couple other small improvements:

1. Use new, safer `tryDispose` in mergeConcurrently.
2. Cancel scheduled task wrapper in ValueSource, instead of only disposing the task itself.  Canceling the scheduled task wrapper removes it from the scheduler *and* disposes the underlying task.

### DIFF
--- a/lib/combinator/mergeConcurrently.js
+++ b/lib/combinator/mergeConcurrently.js
@@ -51,7 +51,7 @@ Outer.prototype._startInner = function(t, stream) {
 
 Outer.prototype.end = function(t, x) {
 	this.active = false;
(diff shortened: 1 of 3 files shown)


briancavalier:
@xgrommx Could you try this branch in with your `spawn` implementation?  The issue you were seeing seems to have been related to delayed events, rather than with concatMap.  Let me know if this branch works for you.  Thanks!

briancavalier:
Confirmed with @xgrommx via gitter that this is working.  Merging.

PR #205, opened 2016-03-08
Simplify fromPromise using awaitPromises+just.

### LINKED ISSUE #201: sample + fromPromise depends on order
It looks like the order of streams passed to `sample` matters if one or more of them are promises. In the second case `forEach` is never called.

Example:

```js
  describe('most.js/sample-test', () => {
    it('works', (done) => {
      const stream = most.just(1);
      const promise = Promise.resolve(2);
      const sample = most.just(3);

      stream
        .sample((...args) => args, most.fromPromise(promise), sample)
        .forEach(value => assert.ok(value));
    });

    it('does not work', (done) => {
      const stream = most.just(1);
      const promise = Promise.resolve(2);
      const sample = most.just(3);

      stream
        .sample((...args) => args, sample, most.fromPromise(promise)) // order has changed here
        .forEach(value => assert.ok(value));
    });
  })
```

Is this supposed to work like that? I’ve tried it with multiple promises as well. It only works if all promises come first. As soon as one stream comes before one of the promises the stream never emits a value like above.

briancavalier:
This stabilizes sample() and the interaction between `fromPromise()` with *already-fulfilled* promises and `just()`.

[Read this](https://gist.github.com/briancavalier/a27ddc247dd118ae5508) for more information.  It's not clear yet whether this is the best long term solution.  However, it seems reasonable for now, and deletes far more code than it adds.

Close #201

### DIFF
--- a/lib/combinator/promises.js
+++ b/lib/combinator/promises.js
@@ -4,6 +4,7 @@
 
 var Stream = require('../Stream');
 var fatal = require('../fatalError');
+var just = require('../source/core').of;
 
 exports.fromPromise = fromPromise;
 exports.awaitPromises = awaitPromises;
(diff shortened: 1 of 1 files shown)


briancavalier:
Gonna merge this, since it causes the desired effect *and* simplifies a big chunk of code.  It may not be the ultimate solution, but we can improve/revisit it if necessary.

PR #212, opened 2016-03-17
Use mostjs multicast

### LINKED ISSUE #209: Small bug when "caching" stream
Given 
```js
const bodyStream = fromEvent('click', document.body);

// use it below

bodyStream
.filter(event => event.target.matches('.clicker'))
.observe(x => log(x.target));

bodyStream
.filter(event => event.target.matches('.another'))
.observe(x => info(x.target));

// has error if you click anything thats not a button (like the body of the page)
bodyStream
.filter(event => !event.target.matches('.another') && !event.target.matches('.clicker'))
.map(event => document.querySelector(event.className))
.observe(x => info(x.target))
.catch(warn);
```

for convenience bin attached http://esnextb.in/?gist=00e261a0771c4fa55f3d

briancavalier:
This replaces the local multicast with `@most/multicast`, which includes the fix for #209 in https://github.com/mostjs/multicast/pull/6

Close #209


briancavalier:
cc @davidchase

davidchase:
awesome, because we can now bump to the latest and greatest version of `most` :tada:

PR #216, opened 2016-03-26
Implement fusion/flattening for merge

briancavalier:
This implements fusion/flattening for adjacent/nested `merge` operations.  This means there is zero performance hit for nested merges.  For example, if a function returns a merged stream to you, and you then merge it with another stream, that structure will be flattened to a single merge.

* `a.merge(b).merge(c)` effectively becomes `merge(a, b, c)`
* `merge(a, merge(b, c))` effectively becomes `merge(a, b, c)`

It works by concatenating the sources arrays of any nested Merge sources, in effect "flattening" nested merge operations into a single merge.

I had to change `combine` so that it doesn't use mergeSources: since combine is *not* associative, it can't be flattened in the same way.

### DIFF
--- a/lib/combinator/combine.js
+++ b/lib/combinator/combine.js
@@ -7,14 +7,13 @@ var transform = require('./transform');
 var core = require('../source/core');
 var Pipe = require('../sink/Pipe');
 var IndexSink = require('../sink/IndexSink');
-var mergeSources = require('./merge').mergeSources;
 var dispose = require('../disposable/dispose');
 var base = require('../base');
 var invoke = require('../invoke');
 
 var hasValue = IndexSink.hasValue;
 
-//var map = base.map;
+var map = base.map;
 var tail = base.tail;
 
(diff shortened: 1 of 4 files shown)


briancavalier:
This also has the interesting effect of making `mergeArray` obsolete.  For example, the following two stream have *exactly* the same structure and performance characteristics now:

```js
import { merge, mergeArray } from 'most'

const merged1 = mergeArray(arrayOfStreams)

const merged2 = arrayOfStreams.reduce((s1, s2) => merge(s1, s2))
```

And you could do the latter with any foldable structure, such as a tree:

```js
const merged3 = treeOfSources.reduce((s1, s2) => merge(s1, s2))
```

We could consider removing mergeArray from the public API at some point.  It'd be a breaking change, obviously, so needs to wait for at least a minor version bump (since we're still < 1.0.0)

TylorS:
:shipit:

Issue #217, opened 2016-03-27
Array iterations are not lazy?


standy:
In my app im getting an array of item ids, and then trying to fetch the items.
There simplified example: http://requirebin.com/?gist=65ee1826b0805e595f60
```js
var most = require('most');

getList$()
	.concatMap(getItem$)
	.forEach(x => console.log('got', x, new Date().toISOString()));

// helpers
function getList$() {
	return most.from([1,2,3,4,5]);
}
function getItem$(id) {
	return most.fromPromise(fakeFetch(id))
		.map(id => 'item:' + id);
}
function fakeFetch(value) {
	console.log('fetch', value);
	return new Promise(resolve => setTimeout(resolve, 1000, value));
}
```

To my wonder all fetch starts right away:
```
fetch 1 (0s)
fetch 2 (0s)
fetch 3 (0s)
fetch 4 (0s)
fetch 5 (0s)
got item:1 (1s)
got item:2 (1s)
got item:3 (1s)
got item:4 (1s)
got item:5 (1s)
```

But i need like that:
```
fetch 1 (0s)
got item:1 (1s)
fetch 2 (1s)
got item:2 (2s)
```

The only workaround i found is to wrap my fetch to another concatMap + most.just
```
getList$()
	.concatMap(id => most.just(id)
		.concatMap(getItem$)
	)
```

This is not an issue, but more a question:
Is there way to make it simpler? Because double concatMap not seems right

briancavalier:
Hi @standy.  Great question.  Most is lazy at the event stream level, but not at the level of the items within the stream.  That is, event streams are inert until observed, but once observed, eagerly evaluate their events.  In your example, the stream created from an array will eagerly produce all of the numbers from the array, and `concatMap` will eagerly call `getItem$` on each number.

You've raised a good point about `concatMap`.  It would be very easy to change it to map items lazily.  I figure some people would find it unintuitive.  On the other hand, you found it unintuitive that `concatMap` *doesn't* map lazily :)  So, I think I'd like to experiment with that idea.  I'll create a PR (or if you'd like to take a shot at it, let me know!) sometime this week so folks can play with it in a branch, and we can see if there are any performance impacts (good or bad).

In the meantime, you could reduce the array using `continueWith`, which is effectively a lazy concat.  Off the top of my head:

```js
import { empty, from, fromPromise } from 'most'

const lazyConcat = (s, x) =>
  s.continueWith(() => getItem$(x))

[...]

PR #218, opened 2016-03-28
Add mergeMapConcurrently

briancavalier:
:warning: Experiment :warning:

Add `mergeMapConcurrently`, which lazily maps and merges.  Convert `concatMap` to use it, which allows `concatMap` to be lazy.

While the end result is *almost* indistinguishable from eager `concatMap`, it does change when the mapping function is called, which can change the timing of I/O, etc.  See the example in #217.  It eagerly fetches all of the items in parallel with the current eager `concatMap`, but fetches them in a non-overlapping sequence when `concatMap` is lazy.

### DIFF
--- a/lib/combinator/mergeConcurrently.js
+++ b/lib/combinator/mergeConcurrently.js
@@ -7,21 +7,32 @@ var dispose = require('../disposable/dispose');
 var LinkedList = require('../LinkedList');
 
 exports.mergeConcurrently = mergeConcurrently;
+exports.mergeMapConcurrently = mergeMapConcurrently;
 
-function mergeConcurrently(concurrency, stream) {
-	return new Stream(new MergeConcurrently(concurrency, stream.source));
+function id(x) {
+	return x;
 }
 
-function MergeConcurrently(concurrency, source) {
+function mergeConcurrently(concurrency, stream) {
+	return mergeMapConcurrently(function(x) { return x }, concurrency, stream);
+}
+
+function mergeMapConcurrently(f, concurrency, stream) {
+	return new Stream(new MergeConcurrently(f, concurrency, stream.source));
+}
+
+function MergeConcurrently(f, concurrency, source) {
+	this.f = f;
 	this.concurrency = concurrency;
 	this.source = source;
 }
 
 MergeConcurrently.prototype.run = function(sink, scheduler) {
-	return new Outer(this.concurrency, this.source, sink, scheduler);
+	return new Outer(this.f, this.concurrency, this.source, sink, scheduler);
 };
 
-function Outer(concurrency, source, sink, scheduler) {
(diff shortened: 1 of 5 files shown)


standy:
Looks great, but I see one test failing (windows, Node v4.4.1):
```
Failure: switch | when input contains many streams | should switch when new stream arrives
  [expect.toEqual] [1, 1, 1, 2, 2, 3, 3, 3, 4, 4] expected to be equal to [1, 1, 1, 2, 2, 2, 3, 3, 3, 4]
```

briancavalier:
Interesting, didn't fail for me locally or in travis!  I'll look into it.

standy:
False alert, found kind of race condition in tests:
```js
it('should switch when new stream arrives', function() {
	var i = 0;
	var s = map(function() {
		return constant(++i, periodic(10));
	}, periodic(25));

	return sequenceEqual([1,1,1,2,2,2,3,3,3,4], take(10, switchLatest(s)));
});
```
On 50th ms both streems based on `periodic(10)` and `periodic(25)` triggers

standy:
Same issue on master, made PR #219

lib/combinator/mergeConcurrently.js
```diff
@@ -7,21 +7,32 @@ var dispose = require('../disposable/dispose');
...
 exports.mergeConcurrently = mergeConcurrently;
+exports.mergeMapConcurrently = mergeMapConcurrently;
+
+function id(x) {
+	return x;
+}
 
 function mergeConcurrently(concurrency, stream) {
-	return new Stream(new MergeConcurrently(concurrency, stream.source));
+	return mergeMapConcurrently(function(x) { return x }, concurrency, stream);
```
campersau: `function(x) { return x; }` can be replaced with `id`.

briancavalier: heh, messy experimental code is messy!  Thanks :)

briancavalier:
@standy Could you try the latest on this branch?  I ported the `switch` tests to use the virtual timer infrastructure, to reduce quirks due to differences in platform timers, poor timer resolution, etc.

briancavalier:
I'm :+1: on this, and we've gotten positive feedback on it from the community on gitter.  So, I think we're a go on this for 0.19.0.  Next step is to review the concatMap unit tests to see if we can verify the laziness of the mapping function.  We should also make a note of the behavior in the docs.

briancavalier:
Rebased on latest master.  Gonna try to add a unit test that (somehow) verifies the laziness, then merge this.

briancavalier:
Ok, added a test and tried to include some info in the docs.  Feedback very welcome, especially on the docs.

TylorS:
Everything looks in order to me :+1:

briancavalier:
Thanks @TylorS!

PR #219, opened 2016-03-28
Fix race condition in swtich test

standy:
This test fails one of 10 cases
```
Failure: switch | when input contains many streams | should switch when new stream arrives
  [expect.toEqual] [1, 1, 1, 2, 2, 3, 3, 3, 4, 4] expected to be equal to [1, 1, 1, 2, 2, 2, 3, 3, 3, 4]
```
On 50th ms both streems based on periodic(10) and periodic(25) triggers, and it happens on 5th element of resulting array
So change 25 to 27 will fix it for first 10 elements

### DIFF
--- a/test/switch-test.js
+++ b/test/switch-test.js
@@ -60,7 +60,7 @@ describe('switch', function() {
 			var i = 0;
 			var s = map(function() {
 				return constant(++i, periodic(10));
-			}, periodic(25));
+			}, periodic(27));
 
 			return sequenceEqual([1,1,1,2,2,2,3,3,3,4], take(10, switchLatest(s)));
 		});


standy:
My mistake, misunderstood the code of test, so it fails even more often then before.
Will try to figure out how to fix it later

briancavalier:
No worries.  Really appreciate your digging into this.  I haven't been able to get it to fail, but I need to try node 4.4.1 to see if maybe there is something specific to some versions of node.

briancavalier:
I just tried node 4.4.1, but on OS X, and all tests still passed.  I haven't found a way to reproduce the test failure yet.

screen shot 2016-03-28 at 9 28 16 pm

It's possible that there's a race condition at 50ms, like you said, and the race favors a different winner under windows than unix-like platforms (maybe due to system call differences?)

standy:
It is easy to reproduce same issue by using 5 and 2 for periodic methods

Not sure if it was race condition, its more like lack of timers accuracy for `setTimeout(25)` vs 3x `setTimeout(10)`. In some cases setTimeout takes more time to execute, and periodic method (which based on setTimeout i believe) can accumulate that error

Something like that (2:real timer took 12.5 ms instead of 10)
```
periodic(25)  -x---------x---------x---------
2              1---1---1---1---1---1---1---1-
2:should                 2---2---2---2---2---
2:real                   2----2----2---2---2-
3                                  3---3---3-
switch:should -1---1---1-2---2---2-3---3---3-
switch:real   -1---1---1-2----2----3---3---3-
```

Anyway test is fixed, issue can be closed, thank you

briancavalier:
> its more like lack of timers accuracy for setTimeout(25) vs 3x setTimeout(10)

Yeah, I'm pretty sure you're right.  Thanks again!

PR #220, opened 2016-03-29
Use virtual timers for switch tests

briancavalier:
Some tests still haven't ported to use virtual time.  This PR ports the `switch` tests to use virtual time, which isn't subject to any JS timer quirks or platform timer differences.


briancavalier:
Possible fix for spurious test failure from #219

PR #221, opened 2016-03-29
Fix most.scan example

campersau:
The description says: `Maintain a sliding window of (up to) 3 values in an array` and not 10


briancavalier:
Thanks!

PR #224, opened 2016-03-29
Fix build script warnings for windows

standy:
Got some messages while runing `npm run build` on windows

1. `Error parsing arguments for flag 'c': 'warnings=false'`  
Windows may use only double quotes in cmd, and unix can use both. So changing to double will make both happy

2. `output.file will be deprecated: Use 'output.filename' instead`  
Used webpack v1.12.14, pretty obvious message

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   ],
   "scripts": {
     "test": "eslint . && buster-test",
-    "build": "webpack --output-library most --output-library-target umd ./most.js --output-file dist/most.js && uglifyjs dist/most.js -c 'warnings=false' -m -o dist/most.min.js",
+    "build": "webpack --output-library most --output-library-target umd ./most.js --output-file dist/most.js && uglifyjs dist/most.js -c \"warnings=false\" -m -o dist/most.min.js",
     "preversion": "npm run build"
   },
   "repository": {
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "eslint": "^1.10.3",
     "transducers-js": "^0.4.174",
     "uglify-js": "^2.4.23",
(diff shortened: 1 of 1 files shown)


briancavalier:
Thanks @standy, sorry about the windows issues.  I wonder if there is an incompatibility between 1.x versions of webpack.  When I run `npm run build` using 1.10.1 on this branch, I get an error:

```
> npm run build  

> most@0.18.7 build /Users/brian/Projects/cujojs/most
> webpack --output-library most --output-library-target umd ./most.js --output-filename dist/most.js && uglifyjs dist/most.js -c "warnings=false" -m -o dist/most.min.js

Hash: beb375f99a89e67306ba
Version: webpack 1.10.1
Time: 13ms
ERROR: can't read file: dist/most.js

... more npm error output here ...
```

We may need to also bump the minimum version of webpack.  When I run `node_modules/.bin/webpack --help`, here is a screenshot of the command line options that it supports.  Unfortunately `--output-filename` isn't listed, only `--output-file`:

screen shot 2016-03-29 at 8 06 32 pm

That's fairly convincing that webpack likely introduced an incompatible command line API change in 1.x at some point.  I don't mind updating to a newer webpack ... could you do that in this PR?

briancavalier:
I tried webpack 1.12.14.  All I did was `npm install --save-dev webpack@^1.12.14`, and then `npm run build` worked again.  Seems like the right thing to go ahead and update the webpack minimum version in this PR :+1:

standy:
Sure! Coming soon

briancavalier:
Awesome, thank you.  I pulled the latest and it worked.  Merging, and will get a release out soon.

PR #226, opened 2016-04-04
Update fromEvent documentation

jfedyczak:
Add information about handling events with more than one argument.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -324,6 +324,8 @@ When passing an EventTarget, you can provide `useCapture` as the 3rd parameter, 
 
 When the stream ends (for example, by using [take](#take), [takeUntil](#until), etc.), it will automatically be disconnected from the event source.  For example, in the case of DOM events, the underlying DOM event listener will be removed automatically.
 
+When source event has more than one argument, all the arguments will be agregated into array in resulting Stream.
+
 Note on EventEmitter: EventEmitters and EventTargets, such as DOM nodes, behave differently in that EventEmitter allows events to be delivered in the same tick as a listener is added.  When using EventEmitter, `most.fromEvent`, will *ensure asynchronous event delivery*, thereby preventing hazards of "maybe sync, maybe async" (aka zalgo) event delivery.
 
 ```js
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -324,6 +324,8 @@ When passing an EventTarget, you can provide `useCapture` as the 3rd parameter,
 
 When the stream ends (for example, by using [take](#take), [takeUntil](#until), etc.), it will automatically be disconnected from the event source.  For example, in the case of DOM events, the underlying DOM event listener will be removed automatically.
 
+When source event has more than one argument, all the arguments will be agregated into array in resulting Stream.
+
```
briancavalier: How about this: Since this array behavior is only supported for EventEmitter (EventTarget is always single-valued), we could refactor the "Note on EventEmitter" section below to have 2 notes.  What do you think of something like this:

**Notes on EventEmitter**

1. When source event has more than one argument ...
2. EventEmitter allows events to be delivered in the same tick ...

docs/api.md
```diff
@@ -324,6 +324,8 @@ When passing an EventTarget, you can provide `useCapture` as the 3rd parameter,
 
 When the stream ends (for example, by using [take](#take), [takeUntil](#until), etc.), it will automatically be disconnected from the event source.  For example, in the case of DOM events, the underlying DOM event listener will be removed automatically.
 
+When source event has more than one argument, all the arguments will be agregated into array in resulting Stream.
```
briancavalier: agregated -> aggregated

jfedyczak:
Does this concern only EventEmitter events?

briancavalier:
Yeah.  `most.fromEvent` supports both EventTarget and EventEmitter.  However, only EventEmitter events can be multi-valued.  EventTarget is always single-valued, since EventListener is always single-valued (See the [whatwg spec for EventTarget and EventListener](https://dom.spec.whatwg.org/#eventtarget))

jfedyczak:
Ok. I've added another commit reflecting your  suggestions.

briancavalier:
Looks great, @jfedyczak.  Thank you!

PR #233, opened 2016-04-14
Improve switch perf

briancavalier:
This PR returns to a custom impl for `switch` to increase perf substantially.  Switch is a common and useful combinator, so the extra code here seems worthwhile.

The simplified impl in master is pleasing and intuitive, but uses *two* of our biggest perf hits: `multicast`, and  `mergeConcurrently`.  Given that, it performs admirably, but this custom impl annihilates it :)

#### Perf comparison

Custom impl:

```
switch 100000 x 1000 streams
most               1156.98 op/s ±  0.96%   (82 samples)
```

Master:

```
switch 100000 x 1000 streams
most                 56.04 op/s ±  7.22%   (65 samples)
```

#### Size comparison

Custom impl:

```
-rw-r--r--  1 brian  staff   11101 Apr 14 08:29 most.min.js.gz
```

Master:

```
-rw-r--r--  1 brian  staff   10933 Apr 14 08:30 most.min.js.gz
```

### DIFF
--- a/lib/combinator/switch.js
+++ b/lib/combinator/switch.js
@@ -3,10 +3,7 @@
 /** @author John Hann */
 
 var Stream = require('../Stream');
-var MulticastSource = require('@most/multicast').MulticastSource;
-var until = require('./timeslice').takeUntil;
-var mergeConcurrently = require('./mergeConcurrently').mergeConcurrently;
-var map = require('./transform').map;
+var dispose = require('../disposable/dispose')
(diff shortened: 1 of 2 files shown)


TylorS:
:shipit:

briancavalier:
Gonna close and reopen to see if that wakes up Travis.

unscriptable:
craziness

PR #235, opened 2016-04-23
Add slice fusion

briancavalier:
This fuses slice.slice to slice.  Because take and skip are both implemented as slice, it also fuses skip.take and take.skip to slice. This allows *any number* of adjacent take, skip, and slice to be fused into a single slice.

#### Todo

- [x] unit tests to verify that fusion math is correct (existing unit tests all still pass)


briancavalier:
Hmmm, math isn't quite right yet ... close though.

TylorS:
Lgtm

PR #236, opened 2016-04-23
Use mostjs/prelude

briancavalier:
Switch to `@most/prelude` and get rid of `lib/base`.

- [x] Make sure it works
- [x] Delete `lib/base` module and replace `require`s


briancavalier:
Goodbye lib/base

lib/combinator/accumulate.js
```diff
@@ -6,6 +6,7 @@ var Stream = require('../Stream');
 var Pipe = require('../sink/Pipe');
 var runSource = require('../runSource');
 var cons = require('./build').cons;
+var Map = require('../fusion/Map');
```
TylorS: Did you mean to add the Map + Scan fusion?

briancavalier: Whoopsee!  Thanks!

PR #238, opened 2016-04-26
Improve slice perf

briancavalier:
The algorithm slice was using was definitely not optimization friendly.  The slice fusion *was* helping, but only about 25%.  By tweaking the algorithm, the fusion has a much larger effect now.  Using the included perf test ...

This PR:

```
skip(n/4) -> take(n/2) 1000000 integers
most                498.39 op/s ±  0.79%   (82 samples)
```

master:

```
skip(n/4) -> take(n/2) 1000000 integers
most                104.89 op/s ±  1.14%   (79 samples)
```

Once again, since take and skip are implemented using slice, all three operations will get the perf boost.

### DIFF
--- a/lib/combinator/slice.js
+++ b/lib/combinator/slice.js
@@ -45,26 +45,27 @@ function slice(start, end, stream) {
 
 function sliceSource(start, end, source) {
 	if(source instanceof Slice) {
-		var s = start + source.skip;
-		var e = Math.min(s + end, source.skip + source.take);
-		return new Slice(s, e, source.source);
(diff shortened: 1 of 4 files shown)


briancavalier:
It turns out that the there's a bug in the logic (`-` instead of `+), which caused slice to process fewer events than it should have.  IOW, this change actually broke slice, and the perf test results are invalid.

There is still some improvement to be had, but it isn't nearly as substantial as I originally thought.

I'll push a fixed update soon.

PR #239, opened 2016-04-26
Remove coallescing filtermap+map fusion

briancavalier:
I can't find a situation where it actually improves performance, and have found a few where it hurts.  More testing is definitely needed, but opening this now so I'm motivated to do it :)


briancavalier:
If we find a reason to add this back later, we can.  I've still not found a situation where it provides any noticeable improvement.

Issue #160, opened 2015-09-22
Implement [Symbol.observable] for ES7 interop compliance


blesh:
Hey @briancavalier! I know you and @jhusain and I talked a few months ago about interop between reactive streaming libraries like this and I'm here to offer a little help bridging that gap.

we've been working hard on the next version of [RxJS](http://github.com/ReactiveX/RxJS), and we're getting closer. Part of that work was to try to adhere to the current ES7 Observable spec by @zenparsing. The part of particular interest there for this library is the `[Symbol.observable]()` method.

### Implementing [Symbol.observable]

Really all that is involved with that is to add a `[Symbol.observable]()` method to your streaming type, and  have it return any object with a `subscribe(observer)` method, and a `[Symbol.observable]()` method.

The benefits here are:

- interop between streaming libraries
- makes your type usable with frameworks looking for that contract (Angular 2 is likely to have support for this, for example)
- eventual interop with JavaScript core whenever ES7 Observable lands.

I think a basic implementation might look like:

```js
[Symbol.observable]() {
  var self = this;
  return {
    subscribe(observer) {
      self.observe(observer.next.bind(observer))
        .then(observer.complete.bind(observer), observer.error.bind(observer));
    }
    [Symbol.observable]() { return this; }
  };
}
```

### Consuming "lowercase-o" observables...

On your end, you could implement something inside of your `from` method to handle any type with a `[Symbol.observable]` method on it. As a guineapig you can use `@reactivex/rxjs` over npm, if you choose. Again it's pretty basic. You'd just call `[Symbol.observable]()` then call `subscribe(observer)` on the returned object, where observer has `next`, `error` and `complete` methods.

### Other things...

I've actively tried to make any monadic-bind-esque operator in RxJS Next support objects that implement `Symbol.observable`. This means, effectively that if you were to implement this, people could jump between Most.js and RxJS using flatMaps and switches, etc.  I'll be reaching out to the Bacon.js community with the same information.

briancavalier:
Hey @blesh, thanks for the update.  On a quick skim, this sounds good.  I'll take a closer look later this week.

blesh:
:+1:

briancavalier:
[...]

PR #240, opened 2016-04-26
feat(Symbol.observable): adds symbol observable interop point

blesh:
This will enable Most to interop with Kefir, RxJS and libraries like Redux

Related #160

@briancavalier a few things:

1. I'm basically *sure* that's not the right place for the code, so I'm not sure how you want to organize it.
2. It took some hackery to get the unsubscription semantic required by the ES observable.
3. I'm a little concerned about possible promise-error-trapping related weirdness here. Errors thrown well down an observable chain during `error` or `complete` handlers may be squelched, because they're wrapped in promise handlers.

### DIFF
--- a/most.js
+++ b/most.js
@@ -7,7 +7,7 @@ var base = require('@most/prelude');
 var core = require('./lib/source/core');
 var from = require('./lib/source/from').from;
 var periodic = require('./lib/source/periodic').periodic;
-
+var $$observable = require('symbol-observable');
 /**
  * Core stream type
  * @type {Stream}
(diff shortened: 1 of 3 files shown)


blesh:
Also, I won't be hurt if this PR is not for you, I'm just trying to be a good citizen.

I strongly believe in interop, and i think it will benefit us all.

cc/ @zenparsing

briancavalier:
Hey @blesh, thanks.  Coincidentally, I created a [gist of something similar](https://gist.github.com/briancavalier/b7c497edfa634d4b3bdd10075c401cdd) just the other day.  I'll take a look at this soon and see what path we can take to interop.

briancavalier:
@blesh There's an impedance mismatch ES7 Observable's imperative `subscribe` API and most's declarative public API.  I'm betting you felt that tension trying to implement this :)  The best way to do it is at the [Source/Sink level](https://github.com/cujojs/most/wiki/Architecture#source), rather than trying to implement it via public API.

Let me know if you're up for giving that a go.  If so, the gist I linked above is probably a good starting point.

blesh:
Thanks, I'll try to look into it soon. Stuck refactoring a large chunk of RxJS at the moment. Lol. Remind me of I don't get back to you!

blesh:
(Sorry for the typos, I'm on my phone)

blesh:
Closing this in favor of #254.

PR #243, opened 2016-04-27
add lazy.js to test/perf

syarul:
Hey,

Got myself tinkering with this and decide to add lazy.js to test/perf, most of the time 2nd to ```most```, but it beat it 4x in ```zip 2 x 100000 integers``` test. I didn't update the readme since I'm running node v5.7.1.

### DIFF
--- a/test/perf/concatMap.js
+++ b/test/perf/concatMap.js
@@ -5,6 +5,7 @@ var rxjs = require('@reactivex/rxjs');
 var kefir = require('kefir');
 var bacon = require('baconjs');
 var lodash = require('lodash');
+var lazy = require('lazy.js')
 
 var runners = require('./runners');
 var kefirFromArray = runners.kefirFromArray;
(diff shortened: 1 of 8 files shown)


briancavalier:
Hey @syarul, thanks.  Cool to see how well lazy.js performs!

I was using lodash and Array mostly as familiar reference points.  The reactive libs are more for dealing with asynchronous events over time rather than for synchronous data processing.  That said, I don't think it hurts to add another data point.

I'm not surprised that lazy.js beats most on zip :)  Asynchronous zip is tricky, and there's probably plenty of room for improving most's implementation!  Since events are async, the reactive libs all have to do buffering and bookkeeping to account for speed differences.

We're about to release 0.19.0, so I'll take a closer look at this after that lands.

syarul:
Sure thing.

briancavalier:
Hey @syarul, sorry for the lag on this.

After much thought and discussion, we've decided that we should remove lodash and Array from the perf test suite.  The main reason is that they fulfill very different use cases.  They're interesting points of reference, but including them can give the wrong impression: that most, rxjs, etc. should be used for synchronous collection processing.

Removing them gives a more apples-to-apples comparison, which is probably more helpful to reactive users.

So, I'll close this and open a new issue to remove lodash and Array.

PR #244, opened 2016-04-27
POC: Extract fusion rules

briancavalier:
⚠️ Do not merge ⚠️ 

This is a proof of concept of extracting fusion rules to a separate "compiler".  Adding fusion rules directly into combinators starts to clutter the code, and distracts from the base implementation.  It might be nice to maintain a separate set of fusion rules that can be applied to a source chain, similar to how GHC's fusion rewrite rules work.

I'm not entirely happy with the implementation here.  It relies on string name matching to avoid module import cycles.  String matching could cause bugs.  I also don't like that the compiler is hardcoded into the Stream constructor.  I'd like to find better approaches for each of those things before even thinking about merging this.
### Open questions
1. How would a 3rd party combinator participate in fusion compilation?
### Todo
- [ ] Find a way to represent matching rules in a way that is more rigorous than string matching.
- [ ] Find a way to separate the compiler from the Stream constructor, or at least allow the compiler or its rules to be configured/enabled/disabled.
- [ ] Investigate applying rules recursively (could cause infinite loops)


unscriptable:
> How would a 3rd party combinator participate in fusion compilation?

The only way I can see (from looking at the code for 5 min) is to export [`compileSource`](https://github.com/cujojs/most/pull/244/files#diff-c17e3152d987338b24e4a567c33473b6R20) instead of `compile`.  Then, if the rules could be augmented by injecting additional rules into Stream... (there's more to it I'm guessing :) ).

> Find a way to represent matching rules in a way that is more rigorous than string matching.

Why are you using strings instead of `instanceof <Constructor>`?  These [all](https://github.com/cujojs/most/pull/244/files#diff-c17e3152d987338b24e4a567c33473b6R8) look like constructor names, anyways.

unscriptable:
o i c. I'm still looking a bit further and it looks like you don't have access to a meaningful constructor where you're defining the rules.

briancavalier:
Great to have some feedback from @unscriptable :)

> export compileSource instead of compile

I think you're on to something here.  I was thinking something similar.  It seems like the key inside is separating rule construction from compileSource.  One way might be to remove the rules from compile.js, export only compileSource, and construct rules in the main most.js module, since it can import the necessary source constructors without creating an import cycle with Stream.

The missing piece in that setup is how/where to actually do the compilation.  I can see two ways, but both involve setting a field on Stream or Stream.prototype:

1. Add the rules (like `Stream.prototype._compilationRules = rules`, yuck!) and let Stream import compileSource
2. Add the "compiler" (like `Stream.prototype.compile = partial(compileSource, rules)`, less yucky, but still), and let Stream just blindly call `this.compile` (it could default to identity function)

This seems better, but overwriting a prototype property is still not great.  Any ideas how to make that better?

briancavalier:
Any further work here should be done in the context of `@most/core`.

PR #249, opened 2016-05-02
Make scan observe stream immediately

### LINKED ISSUE #208: multicast seems to be unreliable
> This issue is based on my comment here: https://github.com/cujojs/most/issues/207#issuecomment-194845872

I have this stream which stops emitting events to **all** subscribers (some still get the event) when I add `multicast`:

```js
most.fromEvent('visibilitychange', document)
  .map(() => !document.hidden)
  .startWith(!document.hidden);
```

I’m currently trying to reproduce a minimal test case for this. Interestingly adding `forEach` to the stream that stops receiving events makes it work again (with `multicast`).

briancavalier:
As discovered in #208, the underlying use of `continueWith` (via `concat` via `startWith`) in scan was delaying observation of the input stream.  This caused a race condition with later combinators that may choose not to emit an event until all their input streams have emitted at least one event, such as `combine` and `sample`.

This changes makes scan *simultaneously* arrange to emit the initial value, and observe the input stream.  Since streams *must* use the scheduler to begin emitting events, this is safe (or, to put it another way, any stream that emits an event synchronously--and thus would win the race with the initial value--is broken and should be fixed).

Fix #208

### DIFF
--- a/lib/combinator/accumulate.js
+++ b/lib/combinator/accumulate.js
@@ -5,7 +5,8 @@
 var Stream = require('../Stream');
 var Pipe = require('../sink/Pipe');
 var runSource = require('../runSource');
-var cons = require('./build').cons;
+var dispose = require('../disposable/dispose');
+var PropagateTask = require('../scheduler/PropagateTask');
 
 exports.scan = scan;
 exports.reduce = reduce;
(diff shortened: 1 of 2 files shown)


briancavalier:
cc @maxhoffmann Here's the cleaned up solution.  I hope to release it a bit later today.

PR #250, opened 2016-05-02
Add thru(), functional-fluent adapter method

briancavalier:
Most's current direction is to provide non-core functional APIs via separate packages, which gets us true modularity without the problems of modifying prototypes.  This PR adds `stream.thru(f)` to allow functional APIs to be used in a fluent way.  For folks who prefer fluent APIs, this allows using those functional APIs with of mixing the two styles.

For example, fluent users have the option of using `thru`:

```js
import hold from `@most/hold`
import { click } from `@most/dom-event`

click(someElement)
  .tap(ev => ev.preventDefault)
  .map(ev => ev.target.value)
  .thru(hold)
  .observer(x => console.log(x))
```

rather than mixing functional and fluent:

```js
import hold from `@most/hold`
import { click } from `@most/dom-event`

hold(click(someElement)
  .tap(ev => ev.preventDefault)
  .map(ev => ev.target.value))
  .observer(x => console.log(x))
```

### Todo

- [x] Docs & examples

### DIFF
--- a/.markdown-doctest-setup.js
+++ b/.markdown-doctest-setup.js
@@ -17,7 +17,8 @@ function noop () {};
 module.exports = {
   require: {
     most: most,
-   'transducers-js': require('transducers-js')
+		'@most/hold': require('@most/hold'),
+		'transducers-js': require('transducers-js')
   },
 
   globals: {
--- a/.markdown-doctest-setup.js
+++ b/.markdown-doctest-setup.js
@@ -17,8 +17,8 @@ function noop () {};
 module.exports = {
   require: {
     most: most,
-		'@most/hold': require('@most/hold'),
-		'transducers-js': require('transducers-js')
+    '@most/hold': require('@most/hold'),
+    'transducers-js': require('transducers-js')
   },
 
   globals: {
--- a/docs/api.md
+++ b/docs/api.md
@@ -49,6 +49,8 @@ most.js API
 	* [during](#during)
 1. Looping
 	* [loop](#loop)
+1. Adapting fluent APIs
+	* [thru](#thru)
 1. Consuming streams
 	* [reduce](#reduce)
 	* [observe](#observe), alias [forEach](#observe)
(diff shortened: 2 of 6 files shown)


briancavalier:
@TylorS @davidchase Can you take a quick look at the docs?  Also, any other thoughts on naming this?  `thru` is a good name ... just want to make sure before we add it to the public API.

davidchase:
:+1: for thru

docs/api.md
```diff
@@ -1066,6 +1066,43 @@ stream.loop(function(values, x) {
...
+Functional APIs allow for the highest degree of modularity via external packages, such as [`@most/hold`](https://github.com/mostjs/hold), *without the risks of modifying prototypes*.
+
+If you prefer using fluent APIs, `thru` allows using those functional APIs in a fluent style.  For example:
+
+```js
+import hold from `@most/hold`
+import { click } from `@most/dom-event`
+
+click(someElement)
+  .tap(ev => ev.preventDefault)
```
TylorS: Shouldn't this be `ev.preventDefault()` ?

TylorS:
I made one tiny comment on the docs but otherwise its good :+1:

docs/api.md
briancavalier: lol, thanks!

.markdown-doctest-setup.js
```diff
@@ -17,7 +17,8 @@ function noop () {};
 module.exports = {
   require: {
     most: most,
-   'transducers-js': require('transducers-js')
+		'@most/hold': require('@most/hold'),
+		'transducers-js': require('transducers-js')
```
briancavalier: ewww, whitespace.  Will fix.

briancavalier:
blargh, old npm :/

This is turning out to be tricky to write a helpful markdown-doctest compatible example for, since the whole point is to use an external library.  Older npm fails because `@most/hold`'s peerDeps need to be updated.  I guess we have to unwind the dep chain a bit more.

briancavalier:
https://github.com/mostjs/hold/pull/9 should address the issue.  Once it's merged, hopefully we can merge this.

briancavalier:
Hey, look at that, travis says ✅

most.js
```diff
@@ -23,6 +23,24 @@ exports.from     = from;
 exports.periodic = periodic;
 
 //-----------------------------------------------------------------------
+// Fluent adapter
+
+var thru = require('./lib/combinator/thru').thru;
+
+exports.thru = thru;
```
briancavalier: I dunno about this ... the whole point is to use thru as a method.  Having it as a function doesn't make much sense.  I'll remove it and if someone sees a reason it should exist, we can easily add it back.

briancavalier:
Ok, I think this is ready to go.  Gonna merge when travis goes green.

PR #252, opened 2016-05-05
Use observation time for just and fromArray

briancavalier:
Previously, these were just using `0` as their `t` value.  It seems tough to me to rigorously (mathematically) justify any value of `t` for something like `just(1)`.  When does 1 "occur"?

**tl;dr;** See lib/source/core.js and lib/source/fromArray.js for the important changes.

So, after some discussion on gitter, it seemed like the right thing to do is to make a *practical* choice about `t`.  We bounced around several ideas:

1. Use observation time, i.e. `scheduler.now()` at the time the Stream provides the value to an observer.
  - There was discussion about what the time origin of `scheduler.now()` should be.  For example, currently it is the same origin as JavaScript `Date`, the unix epoch.  Another option is *program start*: the scheduler would use `Date.now()` at VM startup time as the time origin, and `t` values would start at zero at program start and increase from there. That led to the discussion of *stream-local* time origin ...
2. Use `t=0`, and switch to using stream-local times and use `t=0`--that is, *every* stream (not only `just` and `from`) gets it's own origin at `t=0`.  This is actually similar to 1, but with a per-stream time origin rather than a global time origin.
3. Use construction time, i.e. `scheduler.now()` at the time `just(1)` is called.  One technical hurdle with this is that, at construction time, streams do not know what scheduler they will be asked to use.  They cannot call `scheduler.now()` until they are first observed.

The change to do 1 was trivial, so I figured I'd put it up for discussion.  The change to use program start time would be relatively simple, but I decided not to do it until there's evidence that it's "better" in some way.  The change to do 2 is more involved, but certainly possible.  The change to do 3 would be much more work.

Along the way, I simplified a bunch of code by relying on the fact that ScheduledTask implements Disposable.  We'll probably want to keep those changes even if we don't switch to observation time for `just` and `from`.

### DIFF
--- a/lib/combinator/limit.js
+++ b/lib/combinator/limit.js
@@ -119,7 +119,7 @@ DebounceSink.prototype._clearTimer = function() {
 	if(this.timer === null) {
 		return false;
 	}
-	this.timer.cancel();
+	this.timer.dispose();
 	this.timer = null;
 	return true;
 };
--- a/lib/scheduler/Scheduler.js
+++ b/lib/scheduler/Scheduler.js
@@ -22,12 +22,10 @@ ScheduledTask.prototype.error = function(e) {
 	return this.task.error(this.time, e);
 };
 
-ScheduledTask.prototype.cancel = function() {
+ScheduledTask.prototype.dispose = function() {
 	this.scheduler.cancel(this);
 	return this.task.dispose();
 };
-
-ScheduledTask.prototype.dispose = ScheduledTask.prototype.cancel;
 
 function runTask(task) {
 	try {
--- a/lib/source/ValueSource.js
+++ b/lib/source/ValueSource.js
@@ -12,13 +12,5 @@ function ValueSource(emit, x) {
 }
 
 ValueSource.prototype.run = function(sink, scheduler) {
-	return new ValueProducer(this.emit, this.value, sink, scheduler);
-};
-
-function ValueProducer(emit, x, sink, scheduler) {
-	this.task = scheduler.asap(new PropagateTask(emit, x, sink));
-}
-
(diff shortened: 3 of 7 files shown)


axefrog:
For what it's worth, I think absolute time is fine. "Stream local" time is nice and all, but I think could potentially become complex or have unexpected results when dealing with multiple streams that are combined asynchronously. For example, stream A is active and multicasted, stream B is active and multicasted, but younger than stream A. Stream C is younger again, and combines both stream A and B. Emission times would be weird and unexpected seeing as each stream exists in its own time bubble.

I think it's perfectly ok to have the developer decide on the start time and perform their own subtraction if required.

briancavalier:
Thanks @axefrog.  I can imagine ways to make stream-local work.  Also, [netwire](https://wiki.haskell.org/Netwire) (plenty of info [here](http://hub.darcs.net/ertes/netwire/browse/README.md) uses stream-local, and it's quite a rigorous library.  It's also *very* different, so you won't necessarily find equivalent concepts.  Anyway, it makes me think it's at least interesting enough to continue thinking about stream-local.  Who knows, maybe it will turn out to be a bad idea!

In the near-medium term (which probably means thru 1.x), I agree: shared absolute time is a good option.  It could be what we end up with in the long term, too.  I just want to spend time thinking about it.

briancavalier:
This can probably be considered a breaking change, so we'll need to do a 0.20.0 to publish it.  I *guess* we could call it a bug, but that kinda feels like we're cheating to avoid a minor version bump.

axefrog:
No harm bumping the version, I'm sure. To be fair, you seem to be shooting pretty hard for a 1.0 release anyway, so it won't make too much difference at the end of the day.

PR #253, opened 2016-05-06
Add fantasy-land Extend

briancavalier:
Implement the [fantasy-land Extend](https://github.com/fantasyland/fantasy-land#extend) type.
### Todo
- [ ] Unit tests
- [ ] Docs

### DIFF
--- a/lib/combinator/extend.js
+++ b/lib/combinator/extend.js
@@ -1,0 +1,18 @@
+/** @license MIT License (c) copyright 2010-2016 original author or authors */
+/** @author Brian Cavalier */
+/** @author John Hann */
+
+var Stream = require('../Stream');
+var transform = require('./transform');
+var startWith = require('./build').cons;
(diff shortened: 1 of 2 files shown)


TrySound:
@briancavalier What are the use cases of extend?2

briancavalier:
@TrySound It's not entirely clear to me how useful it is in practice (which is one reason I haven't worked on this PR in a long time).  One kind of thing that extends easily enables is windowing.

For example, using extend, it's trivial to create a higher-order stream whose nested streams represent a 10-event sliding window:

```js
const sliding window = stream.extend(stream => stream.take(10))
```

In general, extend creates a new stream containing all possible versions of the original, and then you can pick and transform the versions you want.

briancavalier:
Closing this.  I think it's still not clear if it's useful in practice. If there's any further work on it, it will happen in `@most/core` or most 2.0.

PR #254, opened 2016-05-09
Add ES7 interop

### LINKED ISSUE #160: Implement [Symbol.observable] for ES7 interop compliance

briancavalier:
@blesh You inspired me :)  I took a crack at adding interop in both directions.

This is not intended to be a complete implementation of the ES7 spec.  The intention is to provide interoperability between most.js and other libraries that provide minimal ES7 observable compliance, which basically means libs that provide and consume types with a `[Symbol.observable]()` that returns an object with a well-behaved `.subscribe()` method.

### Todo

- [x] Unit tests
- [x] Docs

See #240
Close #160


most.js
```diff
@@ -23,6 +24,19 @@ exports.from     = from;
 exports.periodic = periodic;
 
 //-----------------------------------------------------------------------
+// ES7 Observable interop
+
+var subscribe = require('./lib/observable/subscribe').subscribe;
+
+Stream.prototype.subscribe = function(subscriber) {
+	subscribe(subscriber, this);
```
campersau: Would it make sense to return the disposable here?

briancavalier: Oh my, yes! Good catch @campersau.

blesh:
This looks awesome, @briancavalier!

What do you think about having some shared repository or something with interop tests between known reactive libraries?

most.js
```diff
@@ -23,6 +24,19 @@ exports.from     = from;
...
+// ES7 Observable interop
+
+var subscribe = require('./lib/observable/subscribe').subscribe;
+
+Stream.prototype.subscribe = function(subscriber) {
+	return subscribe(subscriber, this);
+};
+
+Stream.prototype[symbolObservable] = function() {
+	return this;
```
blesh: <3

briancavalier:
Thanks @blesh!

> shared repository or something with interop tests

I think it's a great idea.  I would happily donate any/all the tests in this PR that seem useful for a shared interop tests repo.

TylorS:
Love the idea of a shared interop test repo :+1:

briancavalier:
Ok, I think this has a reasonably complete set of unit tests for a first go.  I'll add docs over the next few days, and hopefully this will be ready to go.

briancavalier:
Rebased.  I think this is ready!  Gonna look over it once more and wait for travis.

TylorS:
:+1:

chicoxyzzy:
@blesh @briancavalier 
> shared repository or something with interop tests between known reactive libraries

does it exist?

briancavalier:
@chicoxyzzy As far as I know, it doesn't exist yet.  @blesh may know more, though.

blesh:
No it doesn't exist yet

PR #255, opened 2016-05-10
Update to buster 0.8.0

briancavalier:
It's been out for a few months now, and seems like a transparent update so far.  Would be good to update if it's seamless.


briancavalier:
Travis says ok :)  Merging!

PR #264, opened 2016-05-14
Defend mergeMapConcurrently, continueWith, recoverWith against exceptions

### LINKED ISSUE #260: Most doesn't handle throwing exceptions from chain.
The expected behaviour of the code bellow is one line "This should be reported" but unfortunatelly most hides the error that happened in `chain`.

```javascript
const most = require('most');

most.from([1,2,3])
    .chain(e => {
        throw new Error('This should be reported');
    })
    .drain()
    .catch(err => console.error(err.message));
```

I'm using node v6.1.0 and most 0.19.3.

briancavalier:
If the provided function throws, mergeMapConcurrently, continueWith, and recoverWith all failed to catch the exception.  The result was that mergeMapConcurrently could swallow errors, and that continueWith and recoverWith would cause the exception to be uncatchable (read: crash Node).  All three caused subsequent recoverWith not to work.

This change adds `try/catch` around calls to the user provided functions to catch and propagate errors into the sink chain.

I converted new and touched unit tests to ES6.  In doing so, I found that I had to update to latest eslint, and update .eslintrc to enable es6 support.  I also moved test/errors-test.js to test/combinator/errors-test.js, where it belongs.

Fix #260


TylorS:
:+1:

PR #269, opened 2016-05-22
Refactor, simplify, and possibly rename Scheduler

briancavalier:
The goal of this PR is to break up the current Scheduler further, by separating it into components that have fewer responsibilities and smaller interfaces.  A secondary goal is to find a better name for Scheduler.

So far, this PR extracts a Timeline interface and initial implementation from Scheduler.  This is a good first step: it should be possible to provide a different Timeline implementation (e.g. based on a hash/mod wheel) if we discover one that is "better".

To avoid confusion with scheduler concepts that exists in other reactive libs (e.g. rxjs), it would be good to find a name that is better suited to most's Scheduler.  Some possibilities we've brainstormed on gitter:

- Session - somewhat inspired by [Netwire's Session type](https://hackage.haskell.org/package/netwire-5.0.1/docs/Control-Wire-Session.html)
- Moment - somewhat inspired by [reactive-banana's Moment type](http://hackage.haskell.org/package/reactive-banana-1.1.0.0/docs/Reactive-Banana-Combinators.html#t:Moment), though it is more different than Netwire's Session


briancavalier:
Just a though: Because naming is hard, we could merge this without renaming Scheduler.  Regardless, Timeline still needs a bit more work.

briancavalier:
It's still not clear to me how to improve the separation between Timeline and Scheduler.  It's not quite right in this PR.  However, since neither Scheduler nor Timeline is really a part of the public API yet, we don't really have to try to perfect the interface between the two just yet.

This PR is still an improvement, so I think we should merge it as is.  We can keep thinking about naming and the internal interface.

davidchase:
looks good to ship 👍  small steps

PR #273, opened 2016-05-27
Stop multicasting periodic by default

### LINKED ISSUE #251: Stop multicasting periodic by default
I think I did this back when the scheduler wasn't as efficient as it is now, to reduce setTimeout/setInterval calls--tho, honestly, I can't remember for sure.

However, as pointed out by @axefrog on gitter, it's inconsistent with the rest most.js's current design.  All other streams are not forced-multicast.  That decision is left to the developer.

Now that the scheduler is quite efficient (at most one setTimeout is ever in flight), we could easily remove the multicast from periodic in order to be more consistent.

briancavalier:
Plus minor cleanup. Close #251

### DIFF
--- a/lib/source/periodic.js
+++ b/lib/source/periodic.js
@@ -4,7 +4,6 @@
 
 var Stream = require('../Stream');
 var dispose = require('../disposable/dispose');
-var MulticastSource = require('@most/multicast').MulticastSource;
 var PropagateTask = require('../scheduler/PropagateTask');
 
 exports.periodic = periodic;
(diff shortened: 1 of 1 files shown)


briancavalier:
In the process of doing this, I noticed that [`fromEvent` for the DOM event case also is multicasted by default](https://github.com/cujojs/most/blob/master/lib/source/fromEvent.js#L26).  That was probably done to avoid registering multiple actual dom event handlers, but I doubt that matters in practice.  For consistency, having a strict default of not multicasting is probably better.

briancavalier:
Also [`@most/dom-event` is not multicast](https://github.com/mostjs/dom-event/blob/master/src/dom-event.js) by default.

PR #275, opened 2016-05-28
chore(TS): update typescript definitons

TylorS:
Add subscribe()
export Stream class needed to write 3rd party libraries in TypeScript
Add Source interface
Add Sink and Scheduler, and Disposable interfaces to complete Source interface
Add Task, ScheduledTask to complete Scheduler interface
Add Subscriber and Subscrition interfaces as needed by `subscribe()`

### DIFF
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -23,7 +23,7 @@ declare type CreateGenerator<A> = (...args: Array<any>) => Generator<A|Promise<A
 
 export interface Sink<A> {
   event(time: number, value: A): void;
-  end(time: number, value: A): void;
+  end(time: number, value?: A): void;
   error(time: number, err: Error): void;
 }
 
@@ -47,7 +47,7 @@ export interface Scheduler {
   periodic(task: Task): ScheduledTask;
   schedule(delay: number, period: number, task: Task): ScheduledTask;
   cancel(task: Task): void;
-  cancelAll(predicate: (val: any) => boolean): void;
+  cancelAll(predicate: (task: Task) => boolean): void;
 }
 
 export interface Disposable<A> {
@@ -61,7 +61,7 @@ export interface Source<A> {
 export interface Subscriber<A> {
   next(value: A): void;
   error(err: Error): void;
-  complete(value: A): void;
+  complete(value?: A): void;
 }
 
 export interface Subscription<A> {
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -47,7 +47,7 @@ declare interface Scheduler {
   periodic(task: Task): ScheduledTask;
   schedule(delay: number, period: number, task: Task): ScheduledTask;
   cancel(task: Task): void;
-  cancelAll(predicate: (any) => boolean): void;
+  cancelAll(predicate: (val: any) => boolean): void;
 }
 
 declare interface Disposable<A> {
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -21,12 +21,63 @@ declare interface Iterable<A> {}
 
 declare type CreateGenerator<A> = (...args: Array<any>) => Generator<A|Promise<A>, any, any>;
 
+declare interface Sink<A> {
+  event(time: number, value: A): void;
+  end(time: number, value: A): void;
+  error(time: number, err: Error): void;
+}
+
+declare interface Task {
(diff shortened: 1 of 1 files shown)


TylorS:
cc @AlexGalays

AlexGalays:
Nice!

Only one remark:  thru is already there at the bottom of Stream<A> in master?

TylorS:
Nice catch! I didn't even notice, more worried that vsc didn't yell at me for declaring it twice

TylorS:
Fixed :+1:

AlexGalays:
Awesome.

So you need to create new Streams with specific sources of yours?

TylorS:
Yeah, I want to write most-subject in TypeScript :)
So far this is all I have, and these are the types of things I needed to get started. I'll update this PR if there are anymore interfaces I run into, but I think this covers everything

```typescript
import {Stream, Source} from 'most';

export interface SubjectSource<T> extends Source<T> {
  next: (value: T) => void;
  error: (err: Error) => void;
  complete: (value: T) => void;
}

class Subject<T> extends Stream<T> {
  constructor(source: SubjectSource<T>) {
    super(source);
  }

  next (value: T) {
    (<SubjectSource<T>> this.source).next(value);
  }

  error(err: Error) {
    (<SubjectSource<T>> this.source).error(err);
  }

  complete(value: T) {
    (<SubjectSource<T>> this.source).complete(value);
  }
}
```

type-definitions/most.d.ts
```diff
@@ -21,12 +21,59 @@ declare interface Iterable<A> {}
 
 declare type CreateGenerator<A> = (...args: Array<any>) => Generator<A|Promise<A>, any, any>;
 
+export interface Sink<A> {
+  event(time: number, value: A): void;
+  end(time: number, value: A): void;
```
briancavalier: I think we'll need to make the event and end types different.  For example, `from(arrayOfNumbers)` would require a sink with an event type `number`, and end type `void` (or undefined, or whatever TS uses).  Hmmm, either that or the end value needs to be optional, i.e. `value: ?A` or however TS does that.

What do you think is best?

type-definitions/most.d.ts
```diff
@@ -21,12 +21,59 @@ declare interface Iterable<A> {}
...
+}
+
+export interface Scheduler {
+  now(): number;
+  asap(task: Task): ScheduledTask;
+  delay(task: Task): ScheduledTask;
+  periodic(task: Task): ScheduledTask;
+  schedule(delay: number, period: number, task: Task): ScheduledTask;
+  cancel(task: Task): void;
+  cancelAll(predicate: (val: any) => boolean): void;
```
briancavalier: The predicate takes a Task, so I think this should be `cancelAll(predicate: (task: Task) => boolean): void`

type-definitions/most.d.ts
```diff
@@ -21,12 +21,59 @@ declare interface Iterable<A> {}
...
+}
+
+export interface Source<A> {
+  run (sink: Sink<A>, scheduler: Scheduler): Disposable<A>;
+}
+
+export interface Subscriber<A> {
+  next(value: A): void;
+  error(err: Error): void;
+  complete(value: A): void;
```
briancavalier: See above.  Similarly here, not sure complete should be forced to take an A.

type-definitions/most.d.ts
```diff
@@ -174,6 +221,11 @@ declare interface DisposeFn {
   (): void|Promise<any>;
 }
 
+export class Stream<A> {
+  source: Source<A>;
+  constructor(source: Source<A>);
+}
```
briancavalier: :+1:

briancavalier:
Looks good @TylorS.  See question about end/complete value type.  After we figure that out, I think we can merge this.

type-definitions/most.d.ts
TylorS: Yeah I think it'd probably be best to make it optional :+1:

type-definitions/most.d.ts
TylorS: Good catch!

TylorS:
Updated with the fixes :)

briancavalier:
@TylorS awesome, thanks!  @AlexGalays thanks for the assist!

TylorS:
Thank you, it'd be great to get a patch release for this so I can start making use of this right away :)

PR #278, opened 2016-06-04
Fix example in README.md

MichalZalecki:
10 will logs but after 1 sec of delay. We can zip with periodic to emit each in 1 sec intervals.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ Most.js interoperates seamlessly with ES6 and Promises/A+ promises.  For example
 ```js
 // After 4 seconds, logs 10
 most.from([1, 2, 3, 4])
-	.delay(1000)
+	.zip(v => v, most.periodic(1000))
 	.reduce(function(result, y) {
 		return result + y;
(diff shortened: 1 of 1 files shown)


briancavalier:
Hey @MichalZalecki, thanks for noticing that typo!  There are two ways to "fix" the example.  One is to do something like what you did, which is a really clever use of zip :+1:  The other is to change the comment, [which I just did](https://github.com/cujojs/most/commit/4ce83f8aa6ed0b94946cf29b77e494778c0bbc5d).

I think it'd be cool to incorporate your zip example into the api docs.  Would you be open to adding your example to [the zip section of the API docs](https://github.com/cujojs/most/blob/master/docs/api.md#mostzipf-stream1-stream2---stream)?

MichalZalecki:
Thanks for :+1: Created new PR https://github.com/cujojs/most/pull/280

briancavalier:
Cool, thanks! I'll close this in favor of #280

PR #279, opened 2016-06-05
Remove Promise from typescript definition

AlexGalays:
Hello!

The small local Promise definition embedded in `most.d.ts` seems both unneeded and conflicting with two of the most widely used typescript setups:

- Compiling with ES6 as a target, in which case the typescript compiler automatically add `lib.es6.d.ts` which come with a complete Promise definition.

- Compiling with ES5 as a target and including `es6shim.d.ts` which also come with a comprehensive Promise definition. For instance, Angular 2 do it (can't believe I took them as an example! :D) and it seems to be widely advised on various stackoverflow posts by even typescript compiler contributors.
This is the setup I'm using and removing the definition fixed all errors.

The local Promise definition also has an incompatible interface. So if I use `fetch` that return a `Promise` defined in one of the 2 above files, then `most.fromPromise` won't accept it.

Cheers

@briancavalier   @TylorS

### DIFF
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -1,21 +1,6 @@
 declare type SeedValue<S, V> = { seed: S, value: V };
 declare type TimeValue<V>    = { time: number, value: V };
 
-declare interface Thenable<A> {
-    then<U>(onFulfilled?: (value: A) => U | Thenable<U>, onRejected?: (error: any) => U | Thenable<U>): Thenable<U>;
-    then<U>(onFulfilled?: (value: A) => U | Thenable<U>, onRejected?: (error: any) => void): Thenable<U>;
-    catch<U>(onRejected?: (error: any) => U | Thenable<U>): Thenable<U>;
-}
-
-declare interface Promise<A> {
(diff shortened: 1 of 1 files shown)


briancavalier:
Thanks @AlexGalays.

Is there a way we can provide a bit of guidance for folks that will help them understand this "right" way of doing things?  For example, is someone is relatively new to TS, what can we do to help them?  Maybe we could add a short statement about ES6 "just working", and ES5 needing to add es6shim.d.ts?  Any thoughts?

AlexGalays:
Good idea; is that something you would like to see in the main README under a new section?

TylorS:
Ahh cool, glad to know that could be removed, I never liked that being there anyways :+1:

briancavalier:
>  is that something you would like to see in the main README under a new section?

Yeah, that sounds great.  Can you add it in this PR?

briancavalier:
@AlexGalays Do you know when you might be able to add the info to the README?  We're really close to being able to release 1.0.  This is a really nice cleanup, and I'd love to get it in!

AlexGalays:
Added a typescript paragraph to the README: cujojs/most/pull/282

briancavalier:
Merging this now that #282 is merged.

PR #280, opened 2016-06-05
Add example of zipping stream with periodic to emit in intervals

MichalZalecki:
Suggestion from https://github.com/cujojs/most/pull/278#issuecomment-223829955


briancavalier:
lgtm!  Thanks @MichalZalecki

PR #281, opened 2016-06-10
Remove deprecated create and cycle

briancavalier:
Removing deprecated functions before 1.0: `create` and `cycle`.  Removed code, tests, TS type defs, and docs.

### Todo

- [x] port one unit test that currently relied on create


test/accumulate-test.js
```diff
@@ -1,68 +1,68 @@
...
+
+import { fromArray } from '../lib/source/fromArray'
+import { empty, of as just } from '../lib/source/core'
+
+import FakeDisposeSource from './helper/FakeDisposeSource'
+
+const sentinel = { value: 'sentinel' };
+const other = { value: 'other' };
+
+function endWith (endValue, { source }) {
```
briancavalier: This could be useful in other tests.  I'm ok with being lazy about extracting it, iow waiting until there's a need :)

PR #282, opened 2016-06-11
Add typescript paragraph

AlexGalays:


briancavalier:
Thanks @AlexGalays! I'll have time to look at this later today.

README.md
```diff
@@ -106,6 +106,13 @@ bower install --save most`
 <script src="most/dist/most.js"></script>
 ```
 
+### Typescript support
+
+Most.js works with typescript out of the box as it provides local typings that will be read when you import Most.js in your code. You do not need to manually link an external `d.ts` file in your tsconfig.
```
briancavalier: How about linking "provides local typings" to https://github.com/cujojs/most/blob/master/type-definitions/most.d.ts

README.md
```diff
@@ -106,6 +106,13 @@ bower install --save most`
 <script src="most/dist/most.js"></script>
 ```
 
+### Typescript support
+
+Most.js works with typescript out of the box as it provides local typings that will be read when you import Most.js in your code. You do not need to manually link an external `d.ts` file in your tsconfig.  
+Most.js has a dependency on native Promises so a type definition for Promise must be available in your setup:
```
briancavalier: It seems like there should either be another line break before this sentence to make sure it gets formatted as a new paragraph in markdown, or it should be combined with the previous paragraph.

briancavalier:
Looks great.  Just a couple minor comments.

AlexGalays:
Updated

briancavalier:
Thanks!

Issue #283, opened 2016-06-12
Observable functions?


evilsoft:
Is there any reason to not allow observable function to be used with `most.from`?

From what I read from the spec, I cannot find where the Observable has to be an object. Just that it needs to implement the Observer interface. Being able to observe a function would clear up a lot of boiler plate when working with virtual-DOM implementations that allow state changes on some app state to re-render when that state has been changed.

Yes I know this is very imperative in practice, but by allowing it, the imperative bits could be squirreled away in a lib and allow state changes by making an `update` observable function.

Thinking something like this, please do not mind the nasty mutations and what not, just something to throw together for this issue. Also error handling and making sure we were passed an obserserver are not implemented here:

```javascript
function observeFunc(f) {
  const subs = []
  const done = false

  function subscribe(o) {
    if(subs.indexOf(o) === -1) {
      subs.push(o)
    }

    return {
      closed: () => done,
      unsubscribe() {
        const idx = subs.indexOf(o)
        if(idx !== -1) { subs.splice(idx, 1) }
      }
    }
  }

  function BigBoy(x) {
    const y = f(x)
    subs.forEach(s => s.next(y))
  }

  BigBoy[Symbol.observable] = () => {
    return { subscribe }
  } 

  return BigBoy
}

const x  = observeFunc(x => x)
const x$ = most.from(x)

x$.observe(console.log.bind(console, 'observed x: '))

x(2)
x(10)
```

so if `x` was an updater function we use it as eventHandlers that dispatch the new state through the stream.

This could be accomplished by changing this line: https://github.com/cujojs/most/blob/master/lib/observable/getObservable.js#L11

from:
```javascript
if(o != null && typeof o === 'object')
```

to:
```javascript
if(o != null && (typeof o === 'object' || typeof o === 'function'))
```

If you have no objections to this, I can get over my feelz about tab indentation and the use of semi-colons :stuck_out_tongue_winking_eye:  and submit a PR with this change and update specs.

***Edit*** Forgot to add the ability to pass the observer function in. :anguished: and name to something a bit more appropriate.

briancavalier:
[...]

PR #285, opened 2016-06-14
Allow anything that implements Observable to be returned from getObservable

evilsoft:
## Open the flood gates!

In response to [issue#283](https://github.com/cujojs/most/issues/283), this PR proposes letting any *thing* that implements the Observable interface through the `getObservable` function. Right now it appears this function is only used in the `most.from` method.

The criteria has been changed to reflect the wording in the API docs for `most.from` [here](https://github.com/cujojs/most/blob/master/docs/api.md#mostfrom):

> ... it must have a [Symbol.observable]() method that return an object with a well-behaved .subscribe() method.

*(maybe I should add a spelling correction for return?)*

The following was changed:
* check if truthy `o` has an observable Symbol function (`oSf`) on it.
* remove the unbinding and rebinding of the `oSf` (Did not see the need to do this, will totes put back if ya want it)
* verify that the *thing* returned from the `oSf` provides a `subscribe` method. Pretty sure it gets checked further down the line, but makes sense to reject in `most.from`. This is also an opinion and will also change this back to just checking for an `object` type, if desired.
* Added some additional assertions to provide, falsey and truthy values for possible types.
* updated the is valid spec to account for the need to have the `subscribe` method.

### DIFF
--- a/test/observable/getObservable-test.js
+++ b/test/observable/getObservable-test.js
@@ -9,7 +9,6 @@ describe('getObservable', function() {
 		assert.same(null, getObservable(0));
 		assert.same(null, getObservable(1));
 		assert.same(null, getObservable(undefined));
-		assert.same(null, getObservable(null));
 		assert.same(null, getObservable(''));
 		assert.same(null, getObservable('string'));
 		assert.same(null, getObservable(true));
@@ -20,6 +19,7 @@ describe('getObservable', function() {
 		assert.same(null, getObservable({}));
 		assert.same(null, getObservable(null));
 		assert.same(null, getObservable(function() {}));
+		assert.same(null, getObservable([]));
 	});
 
 	it('should throw TypeError for invalid observable', function() {
--- a/test/observable/getObservable-test.js
+++ b/test/observable/getObservable-test.js
@@ -6,17 +6,20 @@ var getObservable = require('../../lib/observable/getObservable');
 
 describe('getObservable', function() {
 	it('should return null for non-object', function() {
+		assert.same(null, getObservable(0));
 		assert.same(null, getObservable(1));
 		assert.same(null, getObservable(undefined));
 		assert.same(null, getObservable(null));
+		assert.same(null, getObservable(''));
 		assert.same(null, getObservable('string'));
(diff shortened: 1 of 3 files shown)


test/observable/getObservable-test.js
```diff
@@ -6,15 +6,20 @@ var getObservable = require('../../lib/observable/getObservable');
 
 describe('getObservable', function() {
 	it('should return null for non-object', function() {
+		assert.same(null, getObservable(0));
 		assert.same(null, getObservable(1));
 		assert.same(null, getObservable(undefined));
+		assert.same(null, getObservable(null));
```
evilsoft: Well, i guess `null` IS technically an object, so maybe I remove this assertion?

evilsoft: Yep, it is in the next `it` removing...

briancavalier:
Thanks @evilsoft.  I'll have time to look at this more closely tomorrow.

> (maybe I should add a spelling correction for return?)

Good catch. Please do!

> remove the unbinding and rebinding of the oSf

Believe it or not, there's a reason behind this madness.  It's possible for a malicious or buggy getter to "attack" code that accesses it twice.  For example, the getter could return a function the first time, and a non-function the second time.  Promise implementations typically take this extra precaution (accessing a foreign thenable's `then` method only once), and I noticed that [es-observable does the same](https://github.com/zenparsing/es-observable/blob/master/src/Observable.js#L267-L271).  So, unlikely as it seems, it's also not much extra code, so I'd like to keep it.

We should probably add a comment about why it's there, though, since it isn't obvious at all!  Let me know if you want to do that in this PR, or if you want me to do it separately.

evilsoft:
Sounds good, I thought there was a good reason for it so I wanted to draw your attention to that change. I will add it back on this PR as I am the one that caused the damage. 
That will make this simple PR 4 commits tho...should I rebase them into one?

briancavalier:
Looks great, @evilsoft.

> should I rebase them into one?

Nah, no worries.  We can just let github squash and merge them.

Thanks again for this!

PR #286, opened 2016-06-14
chore(documentation) 

nissoh:
remove reference of `create`
reference `create` to https://github.com/mostjs/multicast where needed

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1332,7 +1332,7 @@ stream2:                   -1---2---3---4->
 stream1.zip(add, stream2): -2---4---6---8->
 ```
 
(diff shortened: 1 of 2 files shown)


briancavalier:
Hey @nissoh, thanks for finding these dangling refs!  My initial impression is that we want to refactor the text around them to make everything a bit more clear.  I'll take a closer look tonight.

briancavalier:
@nissoh See #292 for a fix to the TOC and `zip` docs.  We also recently moved recipes to the wiki, and in the process, [removed the reference to `most.create`](https://github.com/cujojs/most/wiki/Event-delegation).

Thanks again for pointing these out.

briancavalier:
Closed by #292

PR #294, opened 2016-06-23
Update symbol-observable 1.x usage

### LINKED ISSUE #293: API change in `symbol-observable` package
## Breakage to the Max

So when the `symbol-observable` package went to 1.x, it was babel-ified and as such, any common JS reference needs to pull from default name space here:
* https://github.com/cujojs/most/blob/master/lib/observable/getObservable.js#L5
* https://github.com/cujojs/most/blob/e96e111b2ce8ea74328eb1f10f5dfdf45aeb91ab/most.js#L10
* https://github.com/cujojs/most/blob/5cdd679ef1ff37498363cfe5f2f895aea3eed571/test/most-test.js#L5
* https://github.com/cujojs/most/blob/93bc203f0f6413a857020c2f2e3d15ffc73e8b03/test/observable/getObservable-test.js#L4

Just needs to be:
```javascript
var symbolObservable = require('symbol-observable').default;
```

Although I am curious why the spec is not failing. Makes me question the validity of that test. 

*some time later...*
The reason they pass is because symbolObservable becomes: `{ '[object Object]': [Function] }`. So that is now valid according to the spec as it associates `Symbol.observable` to be the string `[object Object]`.

I could fix this up as it is simple enough for my meek mind to understand and throw together some valid specs if ya want.

briancavalier:
From #293:

> So when the symbol-observable package went to 1.x, it was babel-ified and as such, any common JS reference needs to pull from default name space here:
> ...
> Just needs to be:
>
> var symbolObservable = require('symbol-observable').default;

Fix #293

### DIFF
--- a/lib/observable/getObservable.js
+++ b/lib/observable/getObservable.js
@@ -2,7 +2,7 @@
 /** @author Brian Cavalier */
 /** @author John Hann */
 
-var symbolObservable = require('symbol-observable');
+var symbolObservable = require('symbol-observable').default;
 
 module.exports = getObservable;
 
--- a/most.js
+++ b/most.js
@@ -7,7 +7,7 @@ var base = require('@most/prelude');
 var core = require('./lib/source/core');
 var from = require('./lib/source/from').from;
 var periodic = require('./lib/source/periodic').periodic;
-var symbolObservable = require('symbol-observable');
+var symbolObservable = require('symbol-observable').default;
 
 /**
  * Core stream type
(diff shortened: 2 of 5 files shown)


evilsoft:
How is this passing in 0.12 on travis? Does `buster` just integrate with babel? automagically?

evilsoft:
Ah. nm. `test/buster.js` FTW.
Sorry about that!! Guess with `import` it would have been borked everywhere.

briancavalier:
No worries.  Yeah, after 1.0, we'll get back to porting to ES6.  We've done a bit of testing, and there seem to be some perf issues with the code babel outputs, so we have to work through those.  I can barely stand working in ES5 :)

PR #295, opened 2016-06-28
Bump to 1.0.0

briancavalier:

PR #297, opened 2016-07-16
Remove node 0.12 from travis

briancavalier:
eslint 3 officially dropped support for node 0.12, so this PR drops node 0.12 from travis.  We *could* just decide not to lint on 0.12 and still run unit tests, but I like the simplicity of just moving on, especially since node 4 is LTS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ node_js:
   - "6"
   - "5"
   - "4"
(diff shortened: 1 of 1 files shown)


briancavalier:
See thread [here](https://github.com/cujojs/most/pull/295#issuecomment-232854890) as well.

PR #301, opened 2016-08-07
Es6 imports exports

TylorS:
Converts all source code to use ES2015 imports/exports and switches from webpack to rollup for its build system.


joshburgess:
Does Rollup have significant advantages over Webpack 2 in terms of tree shaking/dead code elimination?

axefrog:
@joshburgess Found this discussion here; might have some answers to your question: https://github.com/rollup/rollup/issues/219

davidchase:
IMO the shift to rollup aligns with the goals of the current build setup of the @most packages in which case rollup and buble are being used thus has shown some nice improvements over the current setup with webpack/babel I don't know if this adds to the answer for the question but I thought I would try to help :)

briancavalier:
Looking good, @TylorS!  Thank you for working on this.  What do you see as the next steps here?

TylorS:
I think this is good to merge, but after this we are 100% setup to incrementally move the codebase to ES6 + flow

TylorS:
I say incrementally because I think we're going to have to figure what we *can* and *cannot* do with ES6 to keep our performance high.

davidchase:
i was going to start up the incremental pieces to help out, using the one i had prior to this PR but it was easier to start fresh based off this PR then to try to merge my branch with this one

briancavalier:
Awesome work, @TylorS!

Sinewyk:
It's the javascript world: `jsxnext:main` is already being deprecated in favor of simply `module` for the field in package.json to look for es6 modules ^^. For now, you can just duplicate `jsnext:main` as `module` for 100% compat for everything out there.

briancavalier:
Thanks for the pointer, @Sinewyk.  Neither of these is official yet, but since we're supporting `jsnext:main`, I don't really see any harm in adding `module` for future compat.

briancavalier:
See #313

PR #305, opened 2016-08-15
Fixes most.d.ts merge, mergeArray definition

nissoh:
The current definition restricts only single generic type of A. `merge` in most is a stream that can be a stream of multiple value types. i.e. `Stream<A | B>` or `Stream<A | B | C>` etc..

Right now this PR it is limited to to up to 5 types.

Note: Most does not restrict merge with a single parameter(`most.merge(stream)`) in javascript this is just fine. but this will fail in typescript since we typed it to either 2 or more types. i think this is negligible so i didn't bother.

@briancavalier i can provide examples if this is not well understood

### DIFF
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -242,8 +242,15 @@ export function continueWith<A>(f: (a: any) => Stream<A>, s: Stream<A>): Stream<
 export function concatMap<A, B>(f: (a: A) => Stream<B>, s: Stream<A>): Stream<B>;
 export function mergeConcurrently<A>(concurrency: number, s: Stream<Stream<A>>): Stream<A>;
 
-export function merge<A>(...ss: Array<Stream<A>>): Stream<A>;
-export function mergeArray<A>(streams: Array<Stream<A>>): Stream<A>;
+export function merge<A, B>(...arr: Array<A | B>): Stream<A | B>;
+export function merge<A, B, C>(...arr: Array<A | B | C>): Stream<A | B | C>;
+export function merge<A, B, C, D>(...arr: Array<A | B | C | D>): Stream<A | B | C | D>;
+export function merge<A, B, C, D, E>(...arr: Array<A | B | C | D | E>): Stream<A | B | C | D | E>;
+
+export function mergeArray<A, B>(arr: Array<A | B>): Stream<A | B>;
+export function mergeArray<A, B, C>(arr: Array<A | B | C>): Stream<A | B | C>;
+export function mergeArray<A, B, C, D>(arr: Array<A | B | C | D>): Stream<A | B | C | D>;
+export function mergeArray<A, B, C, D, E>(arr: Array<A | B | C | D | E>): Stream<A | B | C | D | E>;
 
 export function combine<A, B, R>(
   fn: (a: A, b: B) => R,
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -242,15 +242,25 @@ export function continueWith<A>(f: (a: any) => Stream<A>, s: Stream<A>): Stream<
 export function concatMap<A, B>(f: (a: A) => Stream<B>, s: Stream<A>): Stream<B>;
 export function mergeConcurrently<A>(concurrency: number, s: Stream<Stream<A>>): Stream<A>;
 
-export function merge<A, B>(...arr: Array<A | B>): Stream<A | B>;
-export function merge<A, B, C>(...arr: Array<A | B | C>): Stream<A | B | C>;
-export function merge<A, B, C, D>(...arr: Array<A | B | C | D>): Stream<A | B | C | D>;
-export function merge<A, B, C, D, E>(...arr: Array<A | B | C | D | E>): Stream<A | B | C | D | E>;
+export interface Merge {
+    <A>(stream: Stream<A>): Stream<A>;
+    <A, B>(stream1: Stream<A>, stream2: Stream<B>): Stream<A | B>;
+    <A, B, C>(stream1: Stream<A>, stream2: Stream<B>, stream3: Stream<C>): Stream<A | B | C>;
+    <A, B, C, D>(stream1: Stream<A>, stream2: Stream<B>, stream3: Stream<C>, stream4: Stream<D>): Stream<A | B | C | D>;
+    <A, B, C, D, E>(stream1: Stream<A>, stream2: Stream<B>, stream3: Stream<C>, stream4: Stream<D>): Stream<A | B | C | D | E>;
+}
 
-export function mergeArray<A, B>(arr: Array<A | B>): Stream<A | B>;
-export function mergeArray<A, B, C>(arr: Array<A | B | C>): Stream<A | B | C>;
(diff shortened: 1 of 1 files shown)


type-definitions/most.d.ts
```diff
@@ -242,8 +242,15 @@ export function continueWith<A>(f: (a: any) => Stream<A>, s: Stream<A>): Stream<
 export function concatMap<A, B>(f: (a: A) => Stream<B>, s: Stream<A>): Stream<B>;
 export function mergeConcurrently<A>(concurrency: number, s: Stream<Stream<A>>): Stream<A>;
 
-export function merge<A>(...ss: Array<Stream<A>>): Stream<A>;
-export function mergeArray<A>(streams: Array<Stream<A>>): Stream<A>;
+export function merge<A, B>(...arr: Array<A | B>): Stream<A | B>;
```
TylorS: I'd probably do this a bit differently

```typescript
export interface Merge {
  <A>(stream: Stream<A>): Stream<A>
  <A, B>(stream1: Stream<A>, stream2: Stream<B>): Stream<A | B>
  <A, B, C>(stream1: Stream<A>, stream2: Stream<B>, stream3: Stream<C>): Stream<A | B | C>
  <A, B, C, D>(stream1: Stream<A>, stream2: Stream<B>, stream3: Stream<C>, stream4: Stream<D>): Stream<A | B | C | D>
}

export const merge: Merge;
```

type-definitions/most.d.ts
```diff
@@ -242,8 +242,15 @@ export function continueWith<A>(f: (a: any) => Stream<A>, s: Stream<A>): Stream<
...
+export function merge<A, B, C>(...arr: Array<A | B | C>): Stream<A | B | C>;
+export function merge<A, B, C, D>(...arr: Array<A | B | C | D>): Stream<A | B | C | D>;
+export function merge<A, B, C, D, E>(...arr: Array<A | B | C | D | E>): Stream<A | B | C | D | E>;
+
+export function mergeArray<A, B>(arr: Array<A | B>): Stream<A | B>;
+export function mergeArray<A, B, C>(arr: Array<A | B | C>): Stream<A | B | C>;
+export function mergeArray<A, B, C, D>(arr: Array<A | B | C | D>): Stream<A | B | C | D>;
+export function mergeArray<A, B, C, D, E>(arr: Array<A | B | C | D | E>): Stream<A | B | C | D | E>;
```
TylorS: Following my other example

```typescript
export interface MergeArray {
  <A>([Stream<A>]): Stream<A>
  <A, B>([Stream<A>, Stream<B>]): Stream<A | B>
  <A, B, C>([Stream<A>, Stream<B>, Stream<C>]): Stream<A | B | C>
  <A, B, C, D>([Stream<A>, Stream<B>, Stream<C>, Stream<D>]): Stream<A | B | C | D>
}

export const mergeArray: MergeArray;
```

type-definitions/most.d.ts
nissoh: Looks good. Maybe we should prefix interfaces with `I` character?
It's a small best practice i've heard that reduces collisions and gives a small hintch that it is an interface

type-definitions/most.d.ts
nissoh: i'm getting a compile error on this one

nissoh:
```typescript
export interface Merge {
    <A>(stream: Stream<A>): Stream<A>;
    <A, B>(stream1: Stream<A>, stream2: Stream<B>): Stream<A | B>;
    <A, B, C>(stream1: Stream<A>, stream2: Stream<B>, stream3: Stream<C>): Stream<A | B | C>;
    <A, B, C, D>(stream1: Stream<A>, stream2: Stream<B>, stream3: Stream<C>, stream4: Stream<D>): Stream<A | B | C | D>;
    <A, B, C, D, E>(stream1: Stream<A>, stream2: Stream<B>, stream3: Stream<C>, stream4: Stream<D>): Stream<A | B | C | D | E>;
}

export const merge: Merge;

export interface MergeArray {
    <A>(arr: [Stream<A>]): Stream<A>;
    <A, B>(arr: [Stream<A>, Stream<B>]): Stream<A | B>;
    <A, B, C>(arr: [Stream<A>, Stream<B>, Stream<C>]): Stream<A | B | C>;
    <A, B, C, D>(arr: [Stream<A>, Stream<B>, Stream<C>, Stream<D>]): Stream<A | B | C | D>;
    <A, B, C, D, E>(arr: [Stream<A>, Stream<B>, Stream<C>, Stream<D>]): Stream<A | B | C | D | E>;
}

export const mergeArray: MergeArray;
````

File: type-definitions/most.d.ts (hunk 1 above)
Comment (TylorS) [inline type-definitions/most.d.ts]: I have not seen any TypeScript projects that prefix everything with the `I` character. I don't find collisions too much an issue since you can do `import { X as Y } from './X'` 

@davidchase @britancavalier Any thoughts ^^

**Comment by briancavalier** [comment]
Hi @nissoh!  

tl;dr I think homogeneity is the right place to start.  However, I'm up for discussing use cases where heterogeneous streams solve problems that would be cumbersome to solve with an Either type.

The decision to make merge homogeneous was intentional.  Heterogeneous events are allowed by pure JS because it's effectively untyped (we could do runtime type checking, but in general, most doesn't do that).  However, if you look at implementations in strongly typed languages like Haskell, for example [Reactive-Banana](https://hackage.haskell.org/package/reactive-banana-1.1.0.1/docs/src/Reactive-Banana-Combinators.html#unionWith), [Netwire](https://hackage.haskell.org/package/netwire-5.0.1/docs/Control-Wire-Event.html#g:4), and others, what you'll find is that the "merge" operation is the associative Monoid or Semigroup combinator, aka `mappend` in Haskell.  They only allow merging homogeneous events.

It would be typical in a strongly typed language to use an `Either` type (or other multivalued, strongly typed, structure) as the event value type when you need heterogeneous values.

Unfortunately, there's a bit of overlap in that most implements fantasy-land, and its semigroup operation is named `concat`.  Most implements both `concat` and `merge`, and while both are valid semigroup implementations, I've come to believe that merge is the more useful of the two.

Again, I think homogeneity is the right place to start.  Have you run into some concrete use cases that we can discuss to see if automatic promotion to Either types, i.e. `A | B`, might be useful?

**Comment by nissoh** [comment]
hey @briancavalier, Thanks for references and a thorough explanation.

Merge should remain homogenous i think i can understand that. my take on this would be very weak but i think that i can understand that dealing with Heterogeneous events will probably increase complexity.

my use case was merging 2 streams into `takeUntil` to destroy that chain based on either of these two events.

**Comment by briancavalier** [comment]
@nissoh Cool, glad that was helpful.

> my use case was merging 2 streams into takeUntil to destroy that chain based on either of these two events.

Ah, that is definitely an interesting use case ... one I hadn't thought about.  In general, maybe there are other use cases where we care about the event time and not about the event value.  Your solution looks good.

Another alternative would be to normalize/discard the event values of the streams you need to merge.  I.e. say you needed to use a variable number of streams with `takeUntil`:

```js
import { constant, mergeArray } from 'most'

// "Normalize" stream value to undefined
const ignoreValue = stream => constant(undefined, stream)

stream.takeUntil(mergeArray(arrayOfStreams.map(ignoreValue)))
```

PR #309, opened 2016-09-01
chore(TS): update typescript definitions

TylorS:
Add typings for most/lib/disposable/*
Add typings for most/lib/scheduler/*
Add typings for most/lib/_.js
Add typings for mmost/lib/sink/_
Update exports to ES2015

### DIFF
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -1,677 +1,788 @@
 declare module "most" {
-  type SeedValue<S, V> = { seed: S, value: V };
-  type TimeValue<V>    = { time: number, value: V };
+	type SeedValue<S, V> = { seed: S, value: V };
+	type TimeValue<V>    = { time: number, value: V };
 
-  // es5 does not have typings for this
-  // use any to allow type-casting
-  type Generator<A, B, C> = any
-  type Iterable = any;
+	// es5 does not have typings for this
+	// use any to allow type-casting
+	type Generator<A, B, C> = any
+	type Iterable = any;
 
-  type CreateGenerator<A> = (...args: Array<any>) => Generator<A|Promise<A>, any, any>;
+	type CreateGenerator<A> = (...args: Array<any>) => Generator<A|Promise<A>, any, any>;
 
-  export interface Sink<A> {
-    event(time: number, value: A): void;
-    end(time: number, value?: A): void;
-    error(time: number, err: Error): void;
-  }
+	export interface Sink<A> {
+		event(time: number, value: A): void;
+		end(time: number, value?: A): void;
+		error(time: number, err: Error): void;
+	}
 
-  export interface Task {
-    run(time: number): void;
-    error(time: number, e: Error): void;
-    dispose(): void;
-  }
+	export interface Task {
+		run(time: number): void;
+		error(time: number, e: Error): void;
+		dispose(): void;
+	}
 
-  export interface ScheduledTask {
-    task: Task;
-    run(): void;
-    error(err: Error): void;
-    dispose(): void;
-  }
+	export interface ScheduledTask {
+		task: Task;
+		run(): void;
+		error(err: Error): void;
+		dispose(): void;
+	}
 
-  export interface Timer {
-    now(): number;
-    setTimer(f: () => any, delayTime: number): Task | number;
-    clearTimer(task: Task | number): void;
-  }
+	export interface Timer {
+		now(): number;
+		setTimer(f: () => any, delayTime: number): Task | number;
+		clearTimer(task: Task | number): void;
+	}
 
-  export interface Timeline {
-    nextArrival(): number;
-    isEmpty(): boolean;
-    add(st: ScheduledTask): void;
-    remove(st: ScheduledTask): boolean;
-    removeAll(predicate: (st: Task) => boolean): void;
-    runTasks(time: number, runTask: (task: Task) => any): void;
-  }
+	export interface Timeline {
+		nextArrival(): number;
+		isEmpty(): boolean;
+		add(st: ScheduledTask): void;
+		remove(st: ScheduledTask): boolean;
+		removeAll(predicate: (st: Task) => boolean): void;
+		runTasks(time: number, runTask: (task: Task) => any): void;
+	}
 
-  export interface Scheduler {
-    now(): number;
-    asap(task: Task): ScheduledTask;
-    delay(delay: number, task: Task): ScheduledTask;
-    periodic(period: number, task: Task): ScheduledTask;
-    schedule(delay: number, period: number, task: Task): ScheduledTask;
-    cancel(task: ScheduledTask): void;
-    cancelAll(predicate: (task: Task) => boolean): void;
-  }
+	export interface Scheduler {
+		now(): number;
+		asap(task: Task): ScheduledTask;
+		delay(delay: number, task: Task): ScheduledTask;
+		periodic(period: number, task: Task): ScheduledTask;
+		schedule(delay: number, period: number, task: Task): ScheduledTask;
+		cancel(task: ScheduledTask): void;
+		cancelAll(predicate: (task: Task) => boolean): void;
+	}
 
-  export interface Disposable<A> {
-    dispose(): void | Promise<A>;
-  }
+	export interface Disposable<A> {
+		dispose(): void | Promise<A>;
+	}
 
-  export interface Source<A> {
-    run (sink: Sink<A>, scheduler: Scheduler): Disposable<A>;
-  }
+	export interface Source<A> {
+		run (sink: Sink<A>, scheduler: Scheduler): Disposable<A>;
+	}
 
-  export interface Subscriber<A> {
-    next(value: A): void;
-    error(err: Error): void;
-    complete(value?: A): void;
-  }
+	export interface Subscriber<A> {
+		next(value: A): void;
+		error(err: Error): void;
+		complete(value?: A): void;
+	}
 
-  export interface Subscription<A> {
-    unsubscribe(): void;
-  }
+	export interface Subscription<A> {
+		unsubscribe(): void;
+	}
 
-  export interface Stream<A> {
-    reduce<B>(f: (b: B, a: A) => B, b: B): Promise<B>;
-    observe(f: (a: A) => any): Promise<any>;
-    forEach(f: (a: A) => any): Promise<any>;
-    drain(): Promise<any>;
-    subscribe(subscriber: Subscriber<A>): Subscription<A>;
+	export interface Stream<A> {
+		reduce<B>(f: (b: B, a: A) => B, b: B): Promise<B>;
+		observe(f: (a: A) => any): Promise<any>;
+		forEach(f: (a: A) => any): Promise<any>;
+		drain(): Promise<any>;
+		subscribe(subscriber: Subscriber<A>): Subscription<A>;
 
-    constant<B>(b: B): Stream<B>;
-    map<B>(f: (a: A) => B): Stream<B>;
-    tap(f: (a: A) => any): Stream<A>;
-    chain<B>(f: (a: A) => Stream<B>): Stream<B>;
-    flatMap<B>(f: (a: A) => Stream<B>): Stream<B>;
+		constant<B>(b: B): Stream<B>;
+		map<B>(f: (a: A) => B): Stream<B>;
+		tap(f: (a: A) => any): Stream<A>;
+		chain<B>(f: (a: A) => Stream<B>): Stream<B>;
+		flatMap<B>(f: (a: A) => Stream<B>): Stream<B>;
 
-    ap<B, C>(fs: Stream<(a: A) => B>): Stream<C>;
+		ap<B, C>(fs: Stream<(a: A) => B>): Stream<C>;
 
-    // Note: Without higher-kinded types, the types for these
-    // cannot be written in a completely safe manner; See https://github.com/Microsoft/TypeScript/issues/1290
-    // For better type safety, consider using most.join/switch/switchLatest with thru
-    join(): A;
-    switch(): A;
-    switchLatest(): A;
+		// Note: Without higher-kinded types, the types for these
+		// cannot be written in a completely safe manner; See https://github.com/Microsoft/TypeScript/issues/1290
+		// For better type safety, consider using most.join/switch/switchLatest with thru
+		join(): A;
+		switch(): A;
+		switchLatest(): A;
 
-    continueWith(f: (a: any) => Stream<A>): Stream<A>;
-    concatMap<B>(f: (a: A) => Stream<B>): Stream<B>;
-    mergeConcurrently<B>(concurrency: number): Stream<B>;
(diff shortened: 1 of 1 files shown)


TylorS:
I was thinking about this, and I think that we should definitely do this. I realized that these typings only affect how to import types not how it ends up importing the library. :D

briancavalier:
@TylorS Yeah, I'm for it.  What's the current status here?  Do we just need to do another round of review before merging, or is there more work to be done?

TylorS:
If any one is up for reviewing it, I think it's in good enough shape.

TylorS:
Actually, immediately after posting that, I realized something. `most/lib` can be defined here as having commonjs modules, and (sadly duplicated)  `most/src` can be defined here with ES modules

TylorS:
Updated to support Es6 and Es5 imports and export

brandonpayton:
Assuming it would help, I'm going to try to get time to review this tomorrow. One thing I noticed tonight is that the latest commit switched indentation from spaces to tabs in the type defs.

TylorS:
Hmmm that wasn't on purpose, but I don't know if that really matters?

briancavalier:
I think the only reason it matters is that the `src` dir is using spaces.  It'd be nice to be consistent.

brandonpayton:
I don't have too much further to go on the review but need to call it a night. I'll get back to this as soon as I can. So far, I have thirteen comments, mostly useful.

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+declare module "most/lib/iterable" {
+	export * from 'most/src/iterable';
+}
+
+declare module "most/src/LinkedList" {
+	class LinkedList<T> {
+		add (x: T): void;
+		remove (x: T): void;
+		isEmpty(): boolean;
+		dispose(): Promise<T>;
```
brandonpayton: This doesn't seem quite right. The implementation can return a Promise with a resolved value of `nothing` or an array.

briancavalier:
Thanks for the very thorough review, @brandonpayton!

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
-export function throwError(e: Error): Stream<any>;
-
-export function multicast<A>(s: Stream<A>): Stream<A>;
+declare module "most" {
+	type SeedValue<S, V> = { seed: S, value: V };
+	type TimeValue<V>    = { time: number, value: V };
+
+	// es5 does not have typings for this
+	// use any to allow type-casting
+	type Generator<A, B, C> = any
```
TylorS: I don't know :)

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+	export interface Sink<A> {
+		event(time: number, value: A): void;
+		end(time: number, value?: A): void;
+		error(time: number, err: Error): void;
+	}
+
+	export interface Task {
+		run(time: number): void;
+		error(time: number, e: Error): void;
+		dispose(): void;
```
TylorS: I don't believe so, the asynchronous dispose stuff is from `Disposable.dispose()` which can return a `Promise`

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+	export interface ScheduledTask {
+		task: Task;
+		run(): void;
+		error(err: Error): void;
+		dispose(): void;
+	}
+
+	export interface Timer {
+		now(): number;
+		setTimer(f: () => any, delayTime: number): Task | number;
```
TylorS: @briancavalier I think we've found ourselves in a tricky situation. `Asap` in src/scheduler/ClockTimer.js as @brandonpayton mentions doesn't satisfy either Task or ScheduledTask interfaces.

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+		clearTimer(task: Task | number): void;
+	}
+
+	export interface Timeline {
+		nextArrival(): number;
+		isEmpty(): boolean;
+		add(st: ScheduledTask): void;
+		remove(st: ScheduledTask): boolean;
+		removeAll(predicate: (st: Task) => boolean): void;
```
TylorS: Correct.

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+		drain(): Promise<any>;
+		subscribe(subscriber: Subscriber<A>): Subscription<A>;
+
+		constant<B>(b: B): Stream<B>;
+		map<B>(f: (a: A) => B): Stream<B>;
+		tap(f: (a: A) => any): Stream<A>;
+		chain<B>(f: (a: A) => Stream<B>): Stream<B>;
+		flatMap<B>(f: (a: A) => Stream<B>): Stream<B>;
+
+		ap<B, C>(fs: Stream<(a: A) => B>): Stream<C>;
```
TylorS: AFAIK this is correct. @briancavalier If you could verify that'd help :). I've never used `.ap` so I'm not 100%, but I think it works as 

```js
const streamOfTypeB = streamOfValuesTypeA.ap(streamOfFunctionsFromAToB)
```

However the type definition is still incorrect

```typescript
ap<B>(fs: Stream<(a: A) => B>): Stream<B>;
```

I think is correct

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+
+		// Note: Without higher-kinded types, the types for these
+		// cannot be written in a completely safe manner; See https://github.com/Microsoft/TypeScript/issues/1290
+		// For better type safety, consider using most.join/switch/switchLatest with thru
+		join(): A;
+		switch(): A;
+		switchLatest(): A;
+
+		continueWith(f: (a: any) => Stream<A>): Stream<A>;
```
TylorS: :+1:

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+		): Stream<R>;
+		zip<B, C, D, E, R>(
+			fn: (a: A, b: B, c: C, d: D, e: E) => R,
+			b: Stream<B>,
+			c: Stream<C>,
+			d: Stream<D>,
+			e: Stream<E>
+		): Stream<R>;
+
+		recoverWith<B>(p: (a: B) => A): Stream<A>;
```
TylorS: :+1:

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+		multicast(): Stream<A>;
+
+		thru<B>(transform: (stream: Stream<A>) => Stream<B>): Stream<B>;
+	}
+
+	interface DisposeFn {
```
TylorS: :+1:

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+
+	export function just<A>(a: A): Stream<A>;
+	export function of<A>(a: A): Stream<A>;
+	export function empty(): Stream<any>;
+	export function never(): Stream<any>;
+	export function from<A>(as: Iterable): Stream<A>;
+	export function periodic<A>(period: number, a?: A): Stream<A>;
+	export function fromEvent(event: string, target: any, useCapture?: boolean): Stream<Event>;
+
+	export function unfold<A, B, S>(f: (seed: S) => SeedValue<S, B|Promise<B>>, seed: S): Stream<B>;
```
TylorS: Correct :D

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+}
+
+declare module "most/src/scheduler/PropagateTask" {
+	import { Task, Sink } from 'most';
+	class PropagateTask<T> implements Task {
+		public active: boolean;
+		constructor(run: (t: number, x: T, sink: Sink<T>) => any, value: T, sink: Sink<T>);
+		static event<T>(x: T, sink: Sink<T>): PropagateTask<T>;
+		static end<T>(x: T, sink: Sink<T>): PropagateTask<T>;
+		static error(err: Error, sink: Sink<any>): PropagateTask<any>;
```
TylorS: I think TypeScript is poor in handling this, but when you go to write anything with this you can never expect any type of value but an error to be propagated, however to satisfy interfaces you will have to write `.event` and `.end` on a `Sink` and I felt it was poor to force any types there.

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+declare module "most/lib/disposable/SettableDisposable" {
+	import SettableDisposable from 'most/src/disposable/SettableDisposable';
+	export = SettableDisposable;
+}
+
+declare module "most/src/disposable/dispose" {
+	import { Disposable, Sink } from 'most';
+	import SettableDisposable = require('most/src/disposable/SettableDisposable');
+
+	export function tryDispose<T> (t: number, disposable: Disposable<T>, sink: Sink<T>): Promise<T> | void;
```
TylorS: :+1:

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+	import fatalError from 'most/src/fatalError';
+	export = fatalError;
+}
+
+declare module "most/src/invoke" {
+	interface InvokeFn {
+		<A, B>(f: (a: A) => B, args: [A]): B;
+		<A, B, C>(f: (a: A, b: B) => C, args: [A, B]): C;
+		<A, B, C, D>(f: (a: A, b: B, c: C) => D, args: [A, B, C]): D;
+		<A, B, C, D, E>(f: (a: A, b: B, c: C, d: D) => E, args: [A, B, C, D]): E;
```
TylorS: Well it is possible to `n` number of arguments as the `default` case of `invoke` falls back to `fn.apply(...)`

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+
+declare module "most/lib/invoke" {
+	import invoke from "most/src/invoke"
+	export = invoke;
+}
+
+declare module "most/src/iterable" {
+	// es5 doesn't have typings for an Iterable
+	// use any to allow type-casting
+	type Iterable = any;
```
TylorS: I'm wondering if it's better to not include it and allow users to include them themselves like we do with Promise

type-definitions/most.d.ts
```diff
@@ -1,385 +1,788 @@
...
+
+	export function isIterable(x: any): boolean;
+	export function getIterator(x: any): Iterable;
```
TylorS: :+1:

type-definitions/most.d.ts
TylorS: :+1:

type-definitions/most.d.ts
TylorS: I mixed up the syntaxes

type-definitions/most.d.ts
briancavalier: Hmm, I think we'll need to look at where Asap instances are used.  I don't think Asap needs to conform to Task or ScheduleTask.  It's mostly used as an opaque "handle" to an asap task used internally by ClockTimer only.  For example, the only thing Scheduler does with Asap instances is: 1) create them by calling timer.setTimer, 2) destroy them by passing them to timer.clearTimer, and 3) compare them to `null`.

type-definitions/most.d.ts
brandonpayton: Hi @TylorS,

Based on [the implementation](https://github.com/cujojs/most/blob/e659d37caef962ad52069486e1b4ccb3c098d016/src/index.js#L190-L198), it is the other way around. The context is a stream of functions, the argument is a stream of values, and the result is a stream of values resulting from applying the latest function to the latest value. The `src/combinator/applicative` module [implements this](https://github.com/cujojs/most/blob/f1f11b460f9a37166f712129c839c8e213cf9059/src/combinator/applicative.js) using `combine`.

I'm not sure what the type definition should be, but I believe we should be modeling the above.

type-definitions/most.d.ts
brandonpayton: Ah, that makes sense. Maybe we should add a short comment explaining why.

type-definitions/most.d.ts
brandonpayton: That seems reasonable to me.

brandonpayton:
Hi @TylorS, I'm sorry it's taken a while to respond to your updates. I took a look at the latest, and it looks good to me with the exception of the following:

1. The `ap` definition still appears incorrect against the implementation. Comment [here](https://github.com/cujojs/most/pull/309#discussion_r88819711).
2. Consider adding a comment explaining why constraints are loosened. Comment [here](https://github.com/cujojs/most/pull/309#discussion_r88820336)
3. Update the Timer interface discussed [here](https://github.com/cujojs/most/pull/309#discussion_r80418727). Perhaps adding a simple Asap handle type would be sufficient.

Thanks!

briancavalier:
@TylorS @brandonpayton What's the current status here?  Master is ahead by a good bit now, so just wondering what the next steps here need to be.  Thanks!

brandonpayton:
@briancavalier I believe we just need updates in response to [these three items](#issuecomment-261828247).

briancavalier:
Thanks @brandonpayton.  I'll do my best to reply to each of those at their original comment site.

type-definitions/most.d.ts
briancavalier: This is confusing because Fantasy Land flipped the `ap` spec.  In the version of the spec (basically v0.3.0) that most.js 1.x supports, `ap`'s context (i.e. `this`) should be a Stream of functions, and it's argument should be a Stream of values.  Unfortunately, I think it's technically impossible to write an accurate type signature for it in TS or flow, which is side-benefit of the reversal in later Fantasy Land versions.

I guess the best we can do is what @brandonpayton suggested:

```ts
export interface Stream<A> {
    ap<B, C>(fs: Stream<B>): Stream<C>;
}
```

In most 2.0.0, we'll support Fantasy Land 2.1, which reverses `ap`, such that its possible to write a type signature for it.

type-definitions/most.d.ts
briancavalier: I'd be ok with adding an opaque Asap handle type as per @brandonpayton's suggestion ([see number 3 here](https://github.com/cujojs/most/pull/309#issuecomment-261828247)), if that indeed solves the problem.

briancavalier:
Looking back at [the 3 remaining items](https://github.com/cujojs/most/pull/309#issuecomment-261828247), I think we've covered them in [mostjs/core](https://github.com/mostjs/core).  Personally, I'd like to focus on getting a 1.0 release of that, and then "fixing" the type definitions here by porting most.js to use mostjs/core and updating most.js type definitions in the process.

How does that sound?

TylorS:
I agree!

brandonpayton:
Sounds good to me!

TylorS:
With `@most/core` v1 released now, I don't see a need to keep this PR open. Further issues with types will likely be better sent in that direction.

Thanks for all the hard work here @brandonpayton, the work we did here definitely did not go to waste. Much of it made it over to `@most/*` packages :)

PR #310, opened 2016-09-06
feat(Stream): add static from method to Stream

TylorS:
Adds from static method to the constructor for ES Observable compatibility


briancavalier:
lgtm, thanks!

PR #332, opened 2016-09-20
most.d.ts recoverWith

nissoh:
recoverWith expects a stream of chained value type. I believe this is a the proper typing


TylorS:
Yes, this would be correct. LGTM

briancavalier:
Good catch @nissoh, thanks!

PR #339, opened 2016-10-06
Add flow types

maxhoffmann:
This PR adds type support for facebook’s `flow` type checker. It copies from the existing typescript definitions and changes the syntax slightly.

Unfortunately using types is more complicated for users than just installing a library via npm. These definitions have to be set up in one of the following ways:
1. copy the file to `<ProjectRoot>/flow-typed/most.js`
2. make a PR to add it to the [`flow-typed`](https://github.com/flowtype/flow-typed/) repository
3. update `.flowconfig` to include this file as lib declarations:

```
[libs]
node_modules/most/flow-typed/most.js
```

More information about third-party type checking can be found in the [official docs](https://flowtype.org/docs/third-party.html).

### DIFF
--- a/flow-typed/most.js
+++ b/flow-typed/most.js
@@ -3,9 +3,6 @@ declare module "most" {
   declare type SeedValue<S, V> = { seed: S, value: V };
   declare type TimeValue<V> = { time: number, value: V };
 
-  declare interface Generator<A, B, C> {}
-  declare interface Iterable<A> {}
-
   declare type CreateGenerator<A> = (...args: Array<any>) => Generator<A|Promise<A>, any, any>;
 
   declare export interface Sink<A> {
@@ -38,7 +35,7 @@ declare module "most" {
   }
 
   declare export interface Disposable<A> {
-    dispose(): void | Promise<A>;
+    dispose(): void | Promise<any>;
   }
 
   declare export interface Source<A> {
--- a/flow-typed/most.js
+++ b/flow-typed/most.js
@@ -1,0 +1,414 @@
+declare module "most" {
+
+  declare type SeedValue<S, V> = { seed: S, value: V };
+  declare type TimeValue<V> = { time: number, value: V };
+
+  declare interface Generator<A, B, C> {}
+  declare interface Iterable<A> {}
+
+  declare type CreateGenerator<A> = (...args: Array<any>) => Generator<A|Promise<A>, any, any>;
+
+  declare export interface Sink<A> {
+    event(time: number, value: A): void;
+    end(time: number, value?: A): void;
+    error(time: number, err: Error): void;
+  }
+
+  declare export interface Task {
+    run(time: number): void;
+    error(time: number, e: Error): void;
+    dispose(): void;
+  }
+
+  declare export interface ScheduledTask {
+    task: Task;
+    run(): void;
+    error(err: Error): void;
+    dispose(): void;
+  }
+
+  declare export interface Scheduler {
+    now(): number;
+    asap(task: Task): ScheduledTask;
+    delay(task: Task): ScheduledTask;
+    periodic(task: Task): ScheduledTask;
+    schedule(delay: number, period: number, task: Task): ScheduledTask;
+    cancel(task: Task): void;
+    cancelAll(predicate: (task: Task) => boolean): void;
+  }
+
+  declare export interface Disposable<A> {
+    dispose(): void | Promise<A>;
+  }
+
+  declare export interface Source<A> {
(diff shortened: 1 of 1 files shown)


TylorS:
Brian will know much more about what going on here, but this gets a big :+1: from me!

briancavalier:
Wow, @maxhoffmann, thank you for taking on this challenge!  I've been actively using flow at work for over a year, and am a big fan.  Flow types have been on the most.js roadmap, so this is great.

In 2 days, I'm leaving for a 2 week vacation.  I'll try to take an initial look before then, but probably won't be able to really dig into it until I get back.

I think another option for distribution is to copy the typedefs alongside the dist/most.js file and name them `most.js.flow`.  I've used that technique on several internal work projects, and so far, it's worked out well.  Of course, we don't have to pick just one of these options :)  Like you said, we can provide them all fairly easily if we want.

maxhoffmann:
Ah cool, I didn’t know the `.flow` file extension works too. I’m happy to do what’s the easiest way when using the library. :)

Just started using flow, so I hope this is good as is. I’ve tried using this file in my project and the type checking seems to work quite well. To be honest, it was just a copy paste job from the typescript file and slightly changing the syntax.

The only definition that I couldn’t get to work is the `Stream` class: https://github.com/cujojs/most/blob/master/type-definitions/most.d.ts#L212 There was an issue on the flow repo that said external type definitions for classes are not supported, but maybe I misread it.

Don’t worry if this is open for a while. No need to rush. :)

briancavalier:
Hey @maxhoffmann, I'm back from vacation.  Existed to start diving into this in the next day or so.

flow-typed/most.js
```diff
@@ -0,0 +1,414 @@
+declare module "most" {
+
+  declare type SeedValue<S, V> = { seed: S, value: V };
+  declare type TimeValue<V> = { time: number, value: V };
+
+  declare interface Generator<A, B, C> {}
+  declare interface Iterable<A> {}
```
briancavalier: These are predefined in flow: https://github.com/facebook/flow/blob/master/lib/core.js#L484

flow-typed/most.js
```diff
@@ -0,0 +1,414 @@
...
+    asap(task: Task): ScheduledTask;
+    delay(task: Task): ScheduledTask;
+    periodic(task: Task): ScheduledTask;
+    schedule(delay: number, period: number, task: Task): ScheduledTask;
+    cancel(task: Task): void;
+    cancelAll(predicate: (task: Task) => boolean): void;
+  }
+
+  declare export interface Disposable<A> {
+    dispose(): void | Promise<A>;
```
briancavalier: I just noticed that the TS definition of Disposable has a parameterized type as well. I wonder if it's providing any value.  @TylorS, do you remember why?

If the type parameter isn't providing any value, we could potentially simplify `dispose()` to `dispose(): void | Promise<any>` or potentially even more strictly: `void | Promise<void>`.

TylorS: There is no good reason, just is what it is. :+1: for simplifying to `Promise<any>` with not template type on Disposable. We should do that in the TS typings as well.

xareelee:
Hello,

How do I include the flowtype of most.js to my project?

> 1. copy the file to <ProjectRoot>/flow-typed/most.js
> 2. make a PR to add it to the flow-typed repository
> 3. update .flowconfig to include this file as lib declarations:
>
> ```
> [libs]
> node_modules/most/flow-typed/most.js
> ```


I did not find `flow-typed/most.js` under `node_modules/most/`.

Also, I tried to use `flow-typed install` to add flowtype `most_vx.x.x.js`. It refers to `most/src/` but I did not find any type annotation in those files.

How do I add and update the flowtype automatically?

briancavalier:
@xareelee most.js does not have flow types yet.  The purpose of this PR is to add them.  If you would like to help by contributing to this PR, that would be great!

briancavalier:
Sorry for letting this go, @maxhoffmann.  The good news is that [@most/core](https://github.com/mostjs/core) has complete API flow type coverage.  Even though it's not yet 1.0, it's very usable, and will be the basis for the next most.js major version bump.

maxhoffmann:
No worries! I’m excited to see types coming to `most`. :)

StreetStrider:
Great. I had been backporting typedefs to my project for a while. Looking forward to use new release. Last time I saw typedefs there was a lack of Stream chain methods. Or maybe I couldn't find it.

TrySound:
@briancavalier Why you don't want to add flow definitions in most@1? My team uses it in production and would be great to have definitions out of the box without waiting for new release.

PR #344, opened 2016-10-19
Fix `Iterable` interface

RamIdeas:
The interface for `Iterable` was empty.

When writing something like `most.from([1,2,3])` the result was typed `Stream<{}>` instead of `Stream<number>`.

This pull request fixes cases like that.

Originally I only accounted for arrays with:

```
declare interface Iterable<A> { [key: number]: A; }
```

but, by definition, `Symbol.iterator` is probably more suitable for _Iterators_.

### DIFF
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -2,7 +2,7 @@ declare type SeedValue<S, V> = { seed: S, value: V };
 declare type TimeValue<V>    = { time: number, value: V };
 
 declare interface Generator<A, B, C> {}
-declare interface Iterable<A> {}
+declare interface Iterable<A> { [Symbol.iterator](): IterableIterator<A>; }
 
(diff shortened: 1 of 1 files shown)


briancavalier:
Thanks @RamIdeas.  @TylorS could you take a look a this?

briancavalier:
Hey @RamIdeas, would you mind rebasing this on latest master?  That should allow Travis to pass.  Once we do that, I think we're probably good to merge, unless @TylorS has any concerns.

Thanks!

TylorS:
:+1:

RamIdeas:
Apologies, I made this change in the Github UI - I didn't think I needed to run eslint for such a trivial change.

It seems an extraneous semicolon was the reason Travis wasn't passing.

I've removed it now - all seems good :)

Do you still need me to rebase or do anything else?

briancavalier:
Thanks, @RamIdeas. I think we're all good now. Merging 😄

PR #350, opened 2016-11-01
Type correction fromEvent most.d.ts

nissoh:
To get the proper type you have to overwrite whole expression. this fix gives the ability to manually cast type that has to satisfy  the constraint of 'Event'.

````typescript
fromEvent('mousemove', document.body) // evaluates type of Stream<Event>

fromEvent<MouseEvent>('mousemove', document.body) // evaluates type of Stream<MouseEvent>

fromEvent<CSSStyleDeclaration>('mousemove', document.body) // error [ts] Type 'CSSStyleDeclaration' does not satisfy the constraint 'Event'
````

### DIFF
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -220,7 +220,7 @@ export function empty(): Stream<any>;
 export function never(): Stream<any>;
 export function from<A>(as: Iterable<A>): Stream<A>;
 export function periodic<A>(period: number, a?: A): Stream<A>;
-export function fromEvent(event: string, target: any, useCapture?: boolean): Stream<Event>;
+export function fromEvent<T extends Event>(event: string, target: any, useCapture?: boolean): Stream<T>;
 
 export function unfold<A, B, S>(f: (seed: S) => SeedValue<S, B|Promise<B>>, seed: S): Stream<B>;
 export function iterate<A>(f: (a: A) => A|Promise<A>, a: A): Stream<A>;

### CODE REVIEW

**Comment by TylorS** [comment]
LGTM :+1:

**Comment by briancavalier** [comment]
This seems really cool, @nissoh.  I think I'm understanding correctly that this change is backward compatible with existing TS code that already uses `fromEvent` (the return type will be `Stream<Event>` in those cases), and folks can opt-in to improved typing if they want.

Is that right?

**Comment by nissoh** [comment]
> this change is backward compatible with existing TS code that already uses fromEvent

Yes, it will result in type of `Event` if you haven't manually casted anything. existing code will have the same typing unless you casted something that does not inherit `Event` class. it will result in type error as i mentioned above.

> folks can opt-in to improved typing if they want.

To get the proper typing you will need to manually cast the result of `fromEvent` like so.
````typescript
const mouseMoveStream: Stream<MouseEvent> = fromEvent('mousemove', document.body) as Stream<MouseEvent> // [ts] Type 'Stream<MouseEvent>' 

----
const mouseMoveStream: Stream<MouseEvent> = fromEvent('mousemove', document.body) // [ts] Type 'Stream<Event>' is not assignable to type 'Stream<MouseEvent>'.
````

now typescript will know how to evaluate `mouseMoveStream` correctly without extra casting.

````typescript
const mouseMoveStream = fromEvent<MouseEvent>('mousemove', document.body) // [ts] Type 'Stream<MouseEvent>' 
````

i'm using typescript version 1.8.10. i'm not sure on which version they added the ability to extend generic types.

**Comment by briancavalier** [comment]
@nissoh Great, thanks for clarifying.  To ask a slightly more focused question:  Will code like this (which I have to believe exists in the wild today) still be valid after this change:

```typescript
const mouseMoveStream = fromEvent('mousemove', document.body) // [ts] Type Stream<Event>
```

@TylorS do you have any idea about TS version compatibility for this?

nissoh:
@briancavalier
prior to this commit i thought that this is possible `fromEvent('keyup', document.body) as Stream<CSSStyleDeclaration> // [ts] Type 'Event' is not comparable to type 'CSSStyleDeclaration'.`

I was wrong about manually casting anything. With that understanding i think that it is completely compatible since you have to cast type that inherits `Event`.  (:

TylorS:
@briancavalier I've seen this done elsewhere, and I believe it's compatible with quite a lot of TS versions.

briancavalier:
Cool.  LGTM!

Thanks again, @nissoh.

PR #354, opened 2016-11-12
Export defaultScheduler and PropagateTask

TylorS:
When writing third-party libraries its not uncommon to need to use a scheduler outside of `Source.run` and to interface with using a Scheduler you will need an implementation of `most.Task` which PropagateTask provides.

Also updates type definitions to reflect the change.


type-definitions/most.d.ts
```diff
@@ -391,3 +391,23 @@ export function recoverWith<A, B>(p: (a: B) => Stream<A>, s: Stream<A>): Stream<
...
+	protected active: boolean;
+
+	constructor (run: (time: number, value: T, sink: Sink<T>) => any, value: T, sink: Sink<T>);
+
+	static event <T> (value: T, sink: Sink<T>): PropagateTask<T>;
+	static error (error: Error, sink: Sink<any>): PropagateTask<any>;
+	static end <T> (value: T, sink: Sink<T>): PropagateTask<T>;
+
+	run(time: number): void;
+  error(time: number, e: Error): void;
```
briancavalier: weird whitespace here.

type-definitions/most.d.ts
```diff
@@ -391,3 +391,23 @@ export function recoverWith<A, B>(p: (a: B) => Stream<A>, s: Stream<A>): Stream<
...
 export function multicast<A>(s: Stream<A>): Stream<A>;
+
+declare const defaultScheduler: Scheduler;
+export { defaultScheduler }
+
+export class PropagateTask<T> implements Task {
+	protected _run: (time: number, value: T, sink: Sink<T>) => any
+	protected value: T;
+	protected sink: Sink<T>;
+	protected active: boolean;
```
briancavalier: Is is possible to omit _run and sink from the public type definition?  Hmmm, maybe it's ok to leave sink.  What do you think?

type-definitions/most.d.ts
TylorS: Fixed!

type-definitions/most.d.ts
briancavalier: Ah right, subclassing.  Makes sense.  Thanks :+1:

PR #356, opened 2016-11-17
Core and Community Packages

davidchase:


TylorS:
LGTM

PACKAGES.md
```diff
@@ -0,0 +1,27 @@
+### Most Core Packages & Examples
+| Name          | Description           
+| ------------- |:-------------:
+| [prelude](https://github.com/mostjs/prelude) | Base functional programming utils for mostjs packages
+| [stream](https://github.com/mostjs/stream)     | Core stream type
```
briancavalier: Let's remove this one for now.  We're not using it yet, so I'd rather not promote it.

PACKAGES.md
```diff
@@ -0,0 +1,27 @@
+### Most Core Packages & Examples
```
briancavalier: Do you think we should try to (sub)group/categorize these by their purpose?  For example, `prelude` is a very different kind of thing than `hold`.  Maybe we should group "tools for most.js contributors" like `prelude` and `package-starter`, and "most.js stream add-ons" or some such?

Does that seem like it'd be helpful?

PACKAGES.md
davidchase: remove `stream` ?

davidchase:
made some adjustments let me know what you guys think

PACKAGES.md
briancavalier: Yeah

briancavalier:
lgtm!  Thanks @davidchase!

Issue #360, opened 2016-11-30
calling unsubscribe in React not unsubscribe observer immediately


jcouyang:
i'm working on a lib [react-most](https://github.com/reactive-react/react-most) to make most works with React, a issue bug me recently, what i'm expecting was that when i call unsubscribe imperatively in `componentWillUnmount()`, it should immediately "unsubscribe" the observer
```js
componentWillUnmount(){
  this.subscriptions.unsubscribe()
}
```
but after the component unmount, observer still get called. this will cause a unmounted component call it's `setState`.

in [componentDidMount](https://github.com/reactive-react/react-most/blob/master/lib/react-most.js#L57), i simply merge all streams and subscribe to setState. [L59](https://github.com/reactive-react/react-most/blob/master/lib/react-most.js#L59) is currently my work around to this issue

in [Todomvc](https://github.com/reactive-react/react-most/blob/master/examples/todomvc) example, the issue will happen when
- delete for example todo item with id 0
- there will be 2 streams react on this change, `mainSection`, and `todoItem`
1. `mainSection` will `setState` to remove item 0 in todo array,
2.  React will call todoItem's `componentWillUnmount`(which will call unsubscribe) then unmount the component
3. todoItem 0 itself will also react on this event, even step 2 already call unmount and `subscriptions.unsubscribe`, observer(setState) still been called on this unmounted component.  what i expected is todoItem 0 should stop calling it's observer in sync when unsubscirbe is imperatively called.

jcouyang:
and this is not a React only issue, a smallest example to reproduce that would be...

1. this works fine, 2 will not print out since it's after unsubscribe in the sequence.
```js
var subscription = most.from([()=>1, subscription.unsubscribe, ()=>2])
  .subscribe({next:x=>console.log(x())})
```

2. but this does not work
```js
var subscription = most.just(1)
  .map(x=>subscription.unsubscribe())
  .map(()=>2)
  .subscribe({next:y=>{console.log(y)}})
```
2 will be print out, `unsubscribe()` should be called before next transform `map(()=>2)`, but imperative unsubscribe seems not work on this case.

briancavalier:
Closed by #361

PR #361, opened 2016-12-01
should noop when it's already disposed

jcouyang:
by simply walk around the codebase

i think this is propbaly the fix for issue #360

by checking `disposable.disposed` before calling the `next` callback

thoughts? @briancavalier

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,6 @@
 root = true
 
 [*]
-indent_style = tab
+indent_style = space
+indent_size = 2
 end_of_line = LF
--- a/src/observable/subscribe.js
+++ b/src/observable/subscribe.js
@@ -26,7 +26,7 @@ export function SubscribeObserver (fatalError, subscriber, disposable) {
 }
 
 SubscribeObserver.prototype.event = function (t, x) {
-  if (typeof this.subscriber.next === 'function') {
+  if (!this.disposable.disposed && typeof this.subscriber.next === 'function') {
     this.subscriber.next(x)
   }
 }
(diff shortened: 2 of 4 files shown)


briancavalier:
Hey @jcouyang!  Thanks for diving into this.

I had just started looking at this last night.  [Not all Disposable types have a `.disposed` property](https://github.com/cujojs/most/blob/master/src/disposable/Disposable.js) (they're only required to have a `.dispose()` method), so in the general case this wouldn't be a safe change.  However, in this specific case, it's guaranteed that SubscribeObserver will always receive a SettableDisposable, which will always have a `.disposed` property.  So, this seems safe ... although, it could potentially be a maintenance issue in the face of later refactoring.

I think it's ok to go with it as a short term fix, though 😄 

I had started by adding a new unit test (based on your simple `map()` + `unsubscribe()` example in #360) to prove that the existing code breaks, expecting then to fix the issue and see the test pass.  So, how about this:  I'll paste my unit test here, and you can incorporate it into your PR.

Here it is.  I had added it as test/observable/unsubscribe-test.js.

```js
import { spec, referee } from 'buster'
const { describe, it } = spec
const { fail, assert } = referee

import { fromArray } from '../../src/source/fromArray'
import { map } from '../../src/combinator/transform'
import { subscribe } from '../../src/observable/subscribe'

describe('unsubscribe', () => {
  it('should prevent observation after unsubscribe', (done) => {
    const f = x => {
      assert.same(1, x)
      subscription.unsubscribe()
      done()
    }

    const subscriber = { next: fail, error: fail, complete: fail }
    let subscription = subscribe(subscriber, map(f, fromArray([1, 2])))
  })
})
```

jcouyang:
cool, thanks @briancavalier

PR #366, opened 2016-12-13
Update deps

briancavalier:
Update all deps except `@most/hold`. Lock `@most/hold` to 1.4.2 to avoid markdown-doctest+babel incompat which is breaking the build.  This is a temporary lock until we decide how to deal with `@most/hold`.

Close #364 
Close #329


briancavalier:
Close #364 
Close #329

PR #369, opened 2016-12-15
Prevent end after end when using subscribe

### LINKED ISSUE #368: Receiving two end signals
most.of(42).takeUntil(most.of(42).delay(1000)).subscribe({
  next: (e) => console.log('next:', e),
  error: (err) => console.log('err:', err),
  complete: () => console.log('end!')
})
next: 42
end!
end!

briancavalier:
Def open to suggestions on how to write a better test for this--see comments in test.

Close #368


campersau:
Does this problem also occur if there are mutliple errors?
Because both methods `event` and `end` check for `!this.disposable.disposed` but `error` doesn't.

briancavalier:
@campersau I debated that.  However, silencing errors is probably worse.  I think we should allow them to surface, even after end (or after error), since silencing them could hide an application bug.

However, it's possible that another option is to make them fatal after end.  Not sure if that's too drastic.  Thoughts?

briancavalier:
@campersau gonna merge this, but please feel free to continue discussing the error-after-end behavior here if you have more thoughts.

PR #370, opened 2016-12-15
feat(count): implement count combinator

TylorS:
As per some previous discussion with @briancavalier  and @davidchase  on a primitive that allows keeping a number representation of the amount of events that a stream has emitted, here is an initial implementation of `count`. This is to serve as a place to continue discussion around such an operator.

If/when agreed upon -- Things that need doing

- [ ] Write documentation
- [ ] Publicly export the combinator
- [ ] Add combinator to the Stream prototype
- [ ] Add to TypeScript type definitions

### DIFF
--- a/src/combinator/count.js
+++ b/src/combinator/count.js
@@ -1,0 +1,36 @@
+import Pipe from '../sink/Pipe'
+import PropagateTask from '../scheduler/PropagateTask'
+import { all } from '../disposable/dispose'
+
+export function count (stream) {
+  return new stream.constructor(new Count(stream.source))
+}
+
+class Count {
+  constructor (source) {
+    this.source = source
+  }
+
+  run (sink, scheduler) {
+    const initial = scheduler.asap(PropagateTask.event(0, sink))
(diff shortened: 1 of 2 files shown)


briancavalier:
What are the pros and cons of this custom implementation vs. something like:

```js
export const count = s => skip(1, scan((x, _) => x + 1, 0, s))
```

(or an equivalent impl using `loop` instead of scan, which would avoid the `skip`)

TylorS:
Potential pros

- Not tied to implementations of other combinators
- Single purpose
- Very simple example of creating combinators

Potential cons

- More code (including tests) to maintain
- *Can* be implemented with current combinators

davidchase:
> export const count = s => skip(1, scan((x, _) => x + 1, 0, s))

That would be sweet to see in the recipes section implemented both ways (scan/loop)

Really shows the power of combining combinators 

thoughts?

briancavalier:
> That would be sweet to see in the recipes section implemented

:+1:

TylorS:
I think we decided a while ago not to do anything with the guy :)

PR #371, opened 2016-12-16
Update perf results for node 6.9.2 and latest libs

briancavalier:
NOTE: I had to remove highland from the `merge` test, as it never finishes on my machine.


unscriptable:
OMG

PR #372, opened 2016-12-16
Add xstream to performance benchmarks

staltz:
Just because.
:)


TylorS:
Thanks @staltz. I know there are some perf tests you have in xstream that we don't have here, do you see any benefits to adding them here?

@briancavalier I'm definitely in favor of this :)

staltz:
> I know there are some perf tests you have in xstream that we don't have here, do you see any benefits to adding them here?

Not necessarily. These tests are primarily to check most.js. I just wanted to add xstream because it's 2nd or 3rd in best performance, so it's nice to have it in the list.

staltz:
Ideally these tests would live in some dedicated repo outside a specific stream lib.

briancavalier:
:+1: I'm in favor as well.  Thanks, @staltz.

PR #373, opened 2016-12-17
Add xstream perf results

briancavalier:
Followon to #372 to add perf results that include xstream.

cc @staltz


TylorS:
:ship: :it:

PR #375, opened 2016-12-19
docs(API): add switchLatest

### LINKED ISSUE #374: Add functional `switchLatest` to API docs
Currently, only the fluent style `switch` is mentioned in the API docs. However, because of a native language conflict, `switch` cannot be exported as a functional API, and the provided substitute `switchLatest` has no mention in the API docs.

Frikki:
Before, only the fluent-style `switch` was mentioned in the API docs.
However, because of a native language conflict, `switch` cannot be
exported as a functional API.

This commit adds point-free substitute `switchLatest` to the API docs.

Closes #374

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -65,7 +65,7 @@ most.js API
 	* [sampleWith](#samplewith)
 	* [zip](#zip)
 1. Combining higher order streams
-	* [switch](#switch), alias [switchLatest](#switch)
+	* [switchLatest](#switchlatest), alias [switch](#switchlatest)
 	* [join](#join)
 	* [mergeConcurrently](#mergeconcurrently)
 1. Awaiting promises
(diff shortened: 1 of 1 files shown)


davidchase:
looks good, see if @briancavalier or @TylorS want to merge it in

docs/api.md
```diff
@@ -1361,7 +1361,7 @@ most.from([1, 2, 3, 4])
 ### switch
 
 ####`stream.switch() -> Stream`
-####`most.switch(stream) -> Stream`
+####`most.switchLatest(stream) -> Stream`
```
briancavalier: Using `most.switch(stream)` is still valid for folks who might do `var most = require('most')` or `import * as most from 'most'`.  Should we keep that in the docs, or should we drop it?  If we drop it, that may open up some followup questions:

* Should we deprecate the function name `switch`?
* Should we show `stream.switchLatest()` (i.e. `Stream.prototype.switchLatest`) here in the docs as well?
* If so, for consistency, should we deprecate `Stream.prototype.switch` and go all-in on the name `switchLatest`?

Thoughts?

TylorS: I'm in favor of `switchLatest` only. We should try to avoid aliases as it leads to confusion when one person uses one name and other uses the other.

In this same context I think we should consider removing other aliases like `flatMapEnd/continueWith` and `flatMapError/recoverWith`,  `flatMap/chain`, `until/since`, and probably some others. We should look to have only 1 way to correctly do some desired behavior.

briancavalier: Sounds like the right direction to me :+1:

In terms of immediate action on this PR, I could see something like:

- Change the TOC to say "switchLatest, alias switch"
- Change `stream.switch()` to `stream.switchLatest` in the heading, examples, and body text
- Add "Alias: stream.switch, most.switch" just below the heading [like we've done in other places](https://github.com/Frikki/most/blob/bee24bb1dae1cad20ffabd8d8efa9e7036c5485f/docs/api.md#chain)

Then, we can take the additional action to deprecate all the aliases we plan to delete for 2.0.  Sound right?

davidchase: 💯   sounds great!

Frikki: @briancavalier If all say, 'Go!', I’ll make these doc changes as you mentioned.

@TylorS @davidchase

Frikki: Interpreting that as a green light.

briancavalier: @Frikki Yes, please do.  Thanks!

Frikki:
I believe that I have accommodated your requests, @briancavalier.

PR #378, opened 2016-12-27
Fix xstream switch perf test

briancavalier:
Also, add switch test to the full perf test run.  ht @davidchase for [finding the mistake](https://github.com/mostjs/core/pull/2#issuecomment-269335352) that was preventing the xstream test from running.


davidchase:
Team work!

davidchase:
:shipit:

PR #395, opened 2017-01-11
Simplify fromIterable, remove tick insertion

briancavalier:
fromIterable was quite old code, from a time when we believed that it was more important to provide concurrency friendliness in the case of large or infinite iterables.  However, we have no evidence that anyone uses most with iterables in that way.  So, the simplest thing to do is simply to plow through the iterable as quickly as possible, like we do for arrays in fromArrays.

If it turns out there is a need for a concurrency-friendly version, we can entertain the idea of adding a new API or separate package for that.

### DIFF
--- a/src/source/fromIterable.js
+++ b/src/source/fromIterable.js
@@ -15,27 +15,16 @@ function IterableSource (iterable) {
 }
 
 IterableSource.prototype.run = function (sink, scheduler) {
-  return new IteratorProducer(getIterator(this.iterable), sink, scheduler)
+  return scheduler.asap(new PropagateTask(runProducer, getIterator(this.iterable), sink))
 }
 
-function IteratorProducer (iterator, sink, scheduler) {
-  this.scheduler = scheduler
-  this.iterator = iterator
-  this.task = new PropagateTask(runProducer, this, sink)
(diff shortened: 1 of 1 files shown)


src/source/fromIterable.js
```diff
@@ -15,27 +15,16 @@ function IterableSource (iterable) {
...
+function runProducer (t, iterator, sink) {
+  let r = iterator.next()
 
-function runProducer (t, producer, sink) {
-  var x = producer.iterator.next()
-  if (x.done) {
-    sink.end(t, x.value)
-  } else {
-    sink.event(t, x.value)
+  while (!r.done && this.active) {
```
TylorS: Where does `this.active` come from?

briancavalier: Yeah, this is slightly confusing.  When a PropagateTask executes the function you provided to its constructor, it [uses itself as the `thisArg`](https://github.com/cujojs/most/blob/master/src/scheduler/PropagateTask.js#L34).  It turned out to be useful to be able to look at the `active` state on several occasions, and for better or for worse (probably worse!), this was a way to "pass" the task itself without adding it to the argument list.  I could def see refactoring that at some point to pass something useful as an argument instead.

PR #399, opened 2017-01-14
chore(PACKAGES.MD) update available community packages

davidchase:


TylorS:
:shipit:

PACKAGES.md
```diff
@@ -7,6 +7,7 @@
 | [create](https://github.com/mostjs/create) | Imperatively push events into a Stream
 | [dom-event](https://github.com/mostjs/dom-event) | Streamlined DOM Events
 | [examples](https://github.com/mostjs/examples) | Most.js examples
+| [core](https://github.com/mostjs/core) | Most.js core event stream
```
briancavalier: Hmmm, should we include this here?  It's not really a package that you would use *with* most, rather it's a package that will soon replace the guts of most.  If you use them together right now, for example, you could fairly easily end up accidentally running two competing schedulers.

Thoughts?

davidchase: yeah maybe should remove it

PR #406, opened 2017-02-03
fix(type-definitions): Expand types accepted by from() to match reality

### LINKED ISSUE #405: most.from typing errors
most.d.ts defines

```typescript
export function from<A>(as: Iterable<A>): Stream<A>;
```

But it's certain (from the JS implementation) that `from` actually supports more types, like array, observable, etc. Currently, TypeScript code that uses most.js will not compile if given e.g. `most.from([10,20,30])`.

briancavalier:
Fix `from` TS type definition to allow accepting the full compliment of types it can coerce.

Fix #405

### DIFF
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -44,6 +44,10 @@ export interface Source<A> {
   run (sink: Sink<A>, scheduler: Scheduler): Disposable<A>;
 }
 
+export interface Observable<A> {
+  subscribe(subscriber: Subscriber<A>): Subscription<A>;
+}
+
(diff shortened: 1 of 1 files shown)


davidchase:
:shipit:

briancavalier:
I think `ArrayLike` is a valid TS type (@TylorS, am I remembering correctly?).  Since `from` accepts things that are Array-like, we should probably include that as well.

briancavalier:
a47bee3 loosens Array to ArrayLike (thanks to @TylorS for discussion via slack).  @davidchase if this still looks good to you, I say we merge it.

PR #408, opened 2017-02-12
fix(fromObservable): allow proper error handling

### LINKED ISSUE #407: Error handling and most.fromObservable
Hi!

While integrating most with `@cycle/time`, I noticed some curious behavior.

```js
import {Observable} from 'rxjs/Rx';

function boom () {
  throw new Error('boom!');
}

const stream = most.from(Observable.of('a')).map(boom);

stream.subscribe({
  next (ev) {},

  error (err) {
    console.log('Handled error');
  },

  complete () {}
});
```

I would expect this code to log 'Handled error'. Instead the error is thrown when it occurs, not caught and passed into the error handler in subscribe.

If `most.from(Observable.of('a'))` is replaced with `most.of('a')` the code functions as expected.

TylorS:
Currently there is no error handling around `fromObservable` and can cause errors to be thrown in place rather than through most's sinks to be handled via `recoverWith` etc.

Fixes #407

### DIFF
--- a/src/observable/fromObservable.js
+++ b/src/observable/fromObservable.js
@@ -4,6 +4,7 @@
 
 import Stream from '../Stream'
 import * as dispose from '../disposable/dispose'
+import { tryEnd, tryEvent } from '../source/tryEvent'
 
 export function fromObservable (observable) {
   return new Stream(new ObservableSource(observable))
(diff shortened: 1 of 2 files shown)


briancavalier:
Right.  The issue here is that Rx's of will propagate the value synchronously (i.e. before it's subscribe() returns).  So, the error is technically thrown from within Rx subscribe()'s call stack.  There are a few possible solutions, but I think introducing tryEvent et al is the right thing :+1:

briancavalier:
Hmmm, there's another potential solution: put a try/catch [around the call to the foreign observable's `subscribe`](https://github.com/cujojs/most/blob/fe716024efd995da34283c438e4c0ade96a8e6b2/src/observable/fromObservable.js#L18).  Maybe like:

```js
var sub
try {
  sub = this.observable.subscribe(new SubscriberSink(sink, scheduler))
} catch(e) {
  sink.error(scheduler.now(), e)
  return dispose.empty()
}
```

I'm not sure yet about the tradeoffs of that approach, but initially, it feels like a "tighter" solution around the actual problem.  Any thoughts?

PR #410, opened 2017-02-22
Remove `extends Event` constraint for `fromEvent`

RamIdeas:
The constraint `T extends Event` doesn't account for node.js EventEmitter's which are clearly [accounted for in the implementation](https://github.com/cujojs/most/blob/7d27cd50afa2cacb6ee39fc62b424c73d61fe9e5/src/source/fromEvent.js#L27-L28).

I've added it as an overload to further clarify that in those situations the `useCapture` parameter is not used.

### DIFF
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -227,6 +227,7 @@ export function never(): Stream<any>;
 export function from<A>(as: ArrayLike<A> | Iterable<A> | Observable<A>): Stream<A>;
 export function periodic<A>(period: number, a?: A): Stream<A>;
 export function fromEvent<T extends Event>(event: string, target: any, useCapture?: boolean): Stream<T>;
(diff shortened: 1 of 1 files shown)


briancavalier:
Thanks @RamIdeas.  Seems right to me, but would like @TylorS to sanity check as well.

TylorS:
:shipit:

RamIdeas:
Haha I don't know if that's good or not!

Quick example, if needed:

``` ts
const socket = ...; // WebSocket, TCP Socket, or any ReadableStream
const data$ = most.fromEvent<Buffer>( 'data', socket );
```

`Buffer` doesn't inherit from `Event` so, without this patch, TypeScript will show errors.

briancavalier:
> Haha I don't know if that's good or not!

😄 it's good ... :shipit: = "ship it"

RamIdeas:
Sweet

PR #411, opened 2017-02-22
Added spreadable overload for `sample`

RamIdeas:
This change adds a spreadable overload for `sample`.

This allows an arbitrary number of streams (more than the 2-5 currently typed) to be sampled, which [matches the implementation](https://github.com/cujojs/most/blob/ef629f5efd05569a0ef3a2eb1b981e01aab29574/docs/api.md#sample).

If you have an array of streams with a common type, you can also spread that into the call and expect typing to flow through.


RamIdeas:
I've messed things up with that second commit.

Let me delete this and try again.

PR #423, opened 2017-03-30
api.md: fix space after h4 header and before code block.

StreetStrider:
In `api.md` where backtick code block is involved inside header, space is missing. Such headers was broken. The PR fixes this.


briancavalier:
Strange.  Seems like GitHub changed something with their markdown rendering.  Thanks for the fix, @StreetStrider!

Issue #427, opened 2017-04-17
until/takeUntil with predicate function


OliverJAsh:
The `until`/`takeUntil` function accepts an input stream, however I was wondering if we could add support for a predicate function? This is different from `takeWhile` because it is inclusive of the first item to return `false`.

For inspiration, have a look at the API for RxJava which has similar options (stream or predicate) for its `takeUntil` operator: http://reactivex.io/RxJava/javadoc/rx/Observable.html#takeUntil(rx.functions.Func1)

briancavalier:
Hi @OliverJAsh. I see, yes, there's a subtle difference around the end.  Do you have a concrete use case for it? If you can share it, that might help us think about how broadly useful this kind of operation is, and whether it belongs in the core or another package.

Thanks!

OliverJAsh:
Hi @briancavalier!

My use case is paginating an API response of ordered items until an item exceeds the range, for example: https://gist.github.com/OliverJAsh/2c327ae63941a237594eed34fe60a47b#file-foo-js-L97

I have an async generating that iterates the pages of Twitter `home_timeline` API. I want take the pages until the end of the user's timeline for today has been found (e.g. `takeUntil(page => page.some(tweet => tweet.created_at < new Date()))`).

briancavalier:
Thanks for the additional info @OliverJAsh.  I see what you mean about inclusive vs. exclusive.  At first glance, it seems possible to use `takeWhile` plus `every`, but that isn't inclusive, so you'd only get all the desired pages if the date aligned exactly with a page boundary.

We're trying to be more strict with types, and avoid runtime type checks, so I'd rather not overload the name `takeUntil`, [even though that name is deprecated](https://github.com/cujojs/most/blob/master/docs/api.md#until) (but not yet removed).  Haskell calls it [dropWhileEnd](https://hackage.haskell.org/package/base-4.9.1.0/docs/Data-List.html#v:dropWhileEnd).

Random thoughts about names: `skipAfter`, `endAfter`, `takeUpto` ... other ideas?

OliverJAsh:
I had a look at `dropWhileEnd` and I don't understand the example well enough to say whether or not it is equivalent to what I have in mind.

There is a similar discussion on the subject of naming in this thread on RxJava: https://github.com/ReactiveX/RxJava/issues/1649

briancavalier:
[...]

PR #429, opened 2017-04-23
adds skipAfter combinator

davidchase:
### Summary

Adds `skipAfter` combinator discussed in #427 

### Todo

- [x] [Documentation](https://github.com/cujojs/most/blob/master/docs/api.md), including examples

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -44,6 +44,7 @@ most.js API
 	* [skip](#skip)
 	* [takeWhile](#takewhile)
 	* [skipWhile](#skipwhile)
+	* [skipAfter](#skipAfter)
 	* [until](#until), alias [takeUntil](#until)
 	* [since](#since), alias [skipUntil](#since)
 	* [during](#during)
@@ -888,7 +889,19 @@ Create a new stream containing all events after `predicate` returns false.
 
 ```
 stream:                 -2-4-5-6-8->
-stream.skipWhile(even): -----5-6-8->
+stream.skipAfter(even): -----5-6-8->
+```
+
+### skipAfter
+
+#### `stream.skipAfter(predicate) -> Stream`
+#### `most.skipAfter(predicate, stream) -> Stream`
+
+Create a new stream containing all events before and including when the `predicate` returns true.
+
+```
+stream:                 -1-2-4-5-6-8->
+stream.skipAfter(even): ---1-2->
 ```
 
 ### until
--- a/docs/api.md
+++ b/docs/api.md
@@ -889,7 +889,7 @@ Create a new stream containing all events after `predicate` returns false.
 
 ```
 stream:                 -2-4-5-6-8->
-stream.skipAfter(even): -----5-6-8->
+stream.skipWhile(even): -----5-6-8->
 ```
 
 ### skipAfter
--- a/docs/api.md
+++ b/docs/api.md
@@ -901,7 +901,7 @@ Create a new stream containing all events before and including when the `predica
 
 ```
 stream:                 -1-2-4-5-6-8->
-stream.skipAfter(even): ---1-2->
+stream.skipAfter(even): ---1-2|
 ```
 
 ### until
--- a/docs/api.md
+++ b/docs/api.md
@@ -44,7 +44,7 @@ most.js API
 	* [skip](#skip)
 	* [takeWhile](#takewhile)
 	* [skipWhile](#skipwhile)
-	* [skipAfter](#skipAfter)
+	* [skipAfter](#skipafter)
 	* [until](#until), alias [takeUntil](#until)
 	* [since](#since), alias [skipUntil](#since)
 	* [during](#during)
--- a/docs/api.md
+++ b/docs/api.md
@@ -900,8 +900,8 @@ stream.skipWhile(even): -----5-6-8->
 Create a new stream containing all events before and including when the `predicate` returns true.
 
 ```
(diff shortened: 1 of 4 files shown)


davidchase:
Guys let me know about the wording, making sure it makes the _most_ sense.

Also should I convert the file over to es6 classes etc? or nah because we may pull in `most/core`

docs/api.md
```diff
@@ -891,6 +892,18 @@ stream:                 -2-4-5-6-8->
...
+### skipAfter
+
+#### `stream.skipAfter(predicate) -> Stream`
+#### `most.skipAfter(predicate, stream) -> Stream`
+
+Create a new stream containing all events before and including when the `predicate` returns true.
+
+```
+stream:                 -1-2-4-5-6-8->
+stream.skipAfter(even): ---1-2->
```
briancavalier: I don't think `skipAfter` will delay the events at all, and it should end immediately after the value where predicate becomes `true`.  Like this:

```
stream:                 -1-2-4-5-6-8->
stream.skipAfter(even): -1-2|
```

src/combinator/slice.js
```diff
@@ -159,3 +159,35 @@ SkipWhileSink.prototype.event = function (t, x) {
...
+function SkipAfterSink (p, sink) {
+  this.p = p
+  this.sink = sink
+  this.skipping = false
+}
+
+SkipAfterSink.prototype.event = function event (t, x) {
+  var p = this.p
+  if (this.skipping) {
+    this.sink.end(t, x)
```
briancavalier: This will wait until the next event before ending.  I think it can end immediately after seeing the first value where `p(x)` is true.  How about:

```js
SkipAfterSink.prototype.event = function event (t, x) {
  var p = this.p
  this.skipping = p(x)
  this.sink.event(t, x)

  if (this.skipping) {
    this.sink.end(t, x)
  }
}
```

We may want to add a `if(this.skipping) return` guard at the beginning of the function (similar to the guard in `TakeWhileSink`) as well.

src/index.js
```diff
@@ -508,6 +508,17 @@ Stream.prototype.skipWhile = function (p) {
   return skipWhile(p, this)
 }
 
+/**
+ * stream:                        -123456789->
+ * skipAfter(x => x === 5, stream): ---12345->
```
briancavalier: Similarly, update docs here to show no delay.

test/slice-test.js
```diff
@@ -78,3 +78,14 @@ describe('skipWhile', function () {
     return expectArray([5, 6, 7, 8, 9], s)
   })
 })
+
+describe('skipAfter', function () {
+  it('should skip all elements after the first one for the condition is true', function () {
+    var a = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+    var s = slice.skipAfter(function (x) {
+      return x > 5
```
briancavalier: Not critical, but personal pref: I think x === 5 is slightly easier to understand here.  "Skip all items after the first event whose value is greater than 5" seems like a bit more cognitive load for the next person who will read or refactor this test.

Two other potential tests: `skipAfter(alwaysFalse, s)` ~= `s` for 1) `empty()`, and 2) a non-empty stream

src/combinator/slice.js
```diff
@@ -159,3 +159,35 @@ SkipWhileSink.prototype.event = function (t, x) {
 
   this.sink.event(t, x)
 }
+
+export function skipAfter (p, stream) {
+  return new stream.constructor(new SkipAfter(p, stream.source))
```
briancavalier: Let's use `new Stream` here *for now*.  Otherwise, `skipAfter` will behave differently than other functions when `stream.constructor !== Stream`.  If we decided to switch to `stream.constructor`, then we can switch the whole library in a single PR.

docs/api.md
```diff
@@ -891,6 +892,18 @@ stream:                 -2-4-5-6-8->
...
+### skipAfter
+
+#### `stream.skipAfter(predicate) -> Stream`
+#### `most.skipAfter(predicate, stream) -> Stream`
+
+Create a new stream containing all events before and including when the `predicate` returns true.
+
+```
+stream:                 -1-2-4-5-6-8->
+stream.skipAfter(even): ---1-2|
```
briancavalier: Couple quick suggestions.  The docs here are showing a delay:

```
stream:                 -1-2-4-5-6-8->
stream.skipAfter(even): ---1-2|
```

Let's align the values to show that there's no delay:

```
stream:                 -1-2-4-5-6-8->
stream.skipAfter(even): -1-2|
```

Someone might wonder why 3 is missing from the example.  The result is the same if we include 3, so might be worth it just to remove a possible wtf (probably rare) for readers.

In which case, we'd end up with:

```
stream:                 -1-2-3-4-5-6-8->
stream.skipAfter(even): -1-2|
```

src/index.js
```diff
@@ -508,6 +508,17 @@ Stream.prototype.skipWhile = function (p) {
   return skipWhile(p, this)
 }
 
+/**
+ * stream:                        -123456789->
+ * skipAfter(x => x === 5, stream): ---12345|
```
briancavalier: This is also showing a delay:

```js
/**
 * stream:                        -123456789->
 * skipAfter(x => x === 5, stream): ---12345|
```

Let's remove the delay and make sure the events are aligned:

```js
/**
 * stream:                          -123456789->
 * skipAfter(x => x === 5, stream): -12345|
```

Issue #445, opened 2017-05-21
The object check on subscribe() is to restricting and doesn't allow functions


far-blue:
### Summary
While RsJS and zen-observable allow the object passed to subscribe() to be an object or a function, most.js specifically checks for typeof object which excludes functions.

### Expected result

subscribe(f) where f is a function object is accepted

### Actual Result

subscribe(f) where f is a function object throws an exception

### Versions

- most.js: 1.3

### Steps to reproduce

1. create a function with a next() method
1. pass it to subscribe()

### Code to reproduce
```
function f () {}
f.next = function (value) {}
most.from([1, 2, 4]).subscribe(f);
```

### Fix
Change the line in subscribe() from:
`if (subscriber == null || typeof subscriber !== 'object') {`

to the style of check zen-observable does:
`if (Object(subscriber) !== subscriber) {`

or just check for a function directly:
`if (subscriber == null || (typeof subscriber !== 'object' && typeof subscriber !== 'function')) {`

### Why allow functions?
I'm actually discussing with the knockoutjs team how they could keep backward compatibility of knockoutjs while providing interoperability with the TC39 Observable spec proposal. Knockout 'observables' are functions which we can extend to support the Observable and Observer specifications but only if an Observer is allowed to be a function. The spec. is not entirely clear but as RxJS and zen-observable appear to allow it and it offers a way for other libraries such as knockout to interop I wanted to request the change.

far-blue:
The knockoutjs discussion is here: https://github.com/knockout/tko/issues/22

briancavalier:
Hey @far-blue, thanks.  It does seem like both [zen-observable](https://github.com/zenparsing/zen-observable/blob/master/src/Observable.js#L85) and [rxjs allow it](https://github.com/ReactiveX/rxjs/blob/master/src/util/toSubscriber.ts), so it seems like most should line up with that.  Would you be up for sending a PR that switches to the `Object(subscriber)` check?

far-blue:
There you go :)

briancavalier:
Thanks @far-blue!  I'll release an update soon.

Fixed by #448

PR #448, opened 2017-05-23
Improve the checking of subscriber objects to match RxJS and zen-observable (#445)

far-blue:
### Summary

Use object wrapping rather than the typeof check to determine if subscriber is valid. Correctly handles objects and functions and is compatible with how RxJS and zen-observable behave as they don’t distinguish between objects and functions. (issue #445)


briancavalier:
Great thanks @far-blue!  Could you [add a test here](https://github.com/cujojs/most/blob/master/test/observable/subscribe-test.js) that covers the function case?

PR #454, opened 2017-06-05
fix(type-definitions): Scheduler method signatures

axefrog:
The type definition `Scheduler.delay(task)` was incorrect. I updated it to `Scheduler.delay(delay, task)`.

### DIFF
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -29,7 +29,7 @@ export interface ScheduledTask {
 export interface Scheduler {
   now(): number;
   asap(task: Task): ScheduledTask;
-  delay(task: Task): ScheduledTask;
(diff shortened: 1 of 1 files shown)


type-definitions/most.d.ts
```diff
@@ -29,7 +29,7 @@ export interface ScheduledTask {
 export interface Scheduler {
   now(): number;
   asap(task: Task): ScheduledTask;
-  delay(task: Task): ScheduledTask;
+  delay(delay: number, task: Task): ScheduledTask;
```
briancavalier: LGTM.  Periodic below also needs a `period: number` param.  Could you add that one as well?

axefrog: Good catch, fixed.

PR #455, opened 2017-06-06
generic parameter default for Disposable<A>

axefrog:
`Disposable<A>` takes a generic type argument which is almost never used in practice. To eliminate the unnecessary specificity in most use cases, I've added a generic parameter default of `any`, which changes the definition to:

```ts
export interface Disposable<A = any> {
  dispose(): void | Promise<A>;
}
```

### DIFF
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -36,7 +36,7 @@ export interface Scheduler {
   cancelAll(predicate: (task: Task) => boolean): void;
 }
 
-export interface Disposable<A> {
+export interface Disposable<A = any> {
   dispose(): void | Promise<A>;
 }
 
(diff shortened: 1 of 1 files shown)


TylorS:
This requires TypeScript 2.3 for all users, I'm not sure if that is the right thing to do. Thoughts?

As an FYI though, It's been removed in `@most/core` to be only `Disposable`

axefrog:
Ah yeah I suppose it could be considered a breaking change.

briancavalier:
It seems like the right thing to do is just wait for most 2.0 (which will be based on `@most/core`).

@axefrog is the current type parameter causing any actual pain at the moment?

axefrog:
No, I've switched to the new version of Most anyway.

briancavalier:
Cool 😄

PR #467, opened 2017-08-01
fix(slice): Reinstate immediate synchronous dispose

### LINKED ISSUE #466: Why does zipping with periodic() hangs?
### Summary

I want to print a stream of increasing numbers every 1s, but the code below hangs Node.js forever:

```javascript
// test.js
const { iterate, periodic } = require('most')

iterate(x => x + 1, 0)
  .take(10)
  .zip(x => x, periodic(1000))
  .forEach(x => console.log(x))
```

while this

```javascript
iterate(x => x + 1, 0)
  .take(10)
  .forEach(x => console.log(x))
```

terminates as expected.

### Expected result

Expected output would be:

```sh
$ node test.js
0
1 # after 1s
2 # after 2s
...
```

### Actual Result

```sh
$ node test.js
0 # <- hangs here and CPU usage goes up
```

### Versions

- most.js: 1.5.0

briancavalier:
### Summary

Reinstate immediate synchronous dispose in `take` and `takeWhile` rather than relying solely on end() to dispose the whole Sink chain.  Use SettableDisposable to deal with cases where source stream may end synchronously before `source.run` returns (and thus `SliceSink`'s `this.disposable` is not yet assigned)

Fix #466

### DIFF
--- a/src/combinator/slice.js
+++ b/src/combinator/slice.js
@@ -2,10 +2,12 @@
 /** @author Brian Cavalier */
 /** @author John Hann */
 
-import Stream from '../Stream'
-import Pipe from '../sink/Pipe'
 import * as core from '../source/core'
+import * as dispose from '../disposable/dispose'
+
 import Map from '../fusion/Map'
+import Pipe from '../sink/Pipe'
+import Stream from '../Stream'
 
 /**
  * @param {number} n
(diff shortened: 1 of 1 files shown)


briancavalier:
After some discussion with the most.js team, we like the alternative in the [fix-466-2 branch](https://github.com/cujojs/most/tree/fix-466-2) better.  Closing this, and will open a new PR on that branch.

PR #472, opened 2017-08-10
Dispose the source disposable

ppoliani:
When disposing a debounce stream we also need to dispose the source disposable

### Summary

I was trying to use `debounce` in the following scenario

```
combine(...streams$)
    .debounce(1000)
    .chain(chainStream)
    .subscribe(observer);
```

But I soon realized that when I tried to dispose the above stream, it would ignore disposing the stream (source stream) that comes before the debounce.
### Todo

- [ ] [Documentation](https://github.com/cujojs/most/blob/master/docs/api.md), including examples

### DIFF
--- a/src/combinator/limit.js
+++ b/src/combinator/limit.js
@@ -4,7 +4,6 @@
 
 import Stream from '../Stream'
 import Pipe from '../sink/Pipe'
-import * as dispose from '../disposable/dispose'
 import PropagateTask from '../scheduler/PropagateTask'
 import Map from '../fusion/Map'
 
--- a/src/combinator/limit.js
+++ b/src/combinator/limit.js
@@ -110,7 +110,7 @@ DebounceSink.prototype.error = function (t, x) {
 
 DebounceSink.prototype.dispose = function () {
   this._clearTimer()
-  this.disposable.dispose()
+  return this.disposable.dispose()
 }
 
 DebounceSink.prototype._clearTimer = function () {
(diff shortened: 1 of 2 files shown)


src/combinator/limit.js
```diff
@@ -110,6 +110,7 @@ DebounceSink.prototype.error = function (t, x) {
 
 DebounceSink.prototype.dispose = function () {
   this._clearTimer()
+  this.disposable.dispose()
```
briancavalier: Yikes.  Good catch, @ppoliani.  This just needs a return, `return this.disposable.dispose()` since disposables typically form a promise chain in the current version of most.js.

briancavalier:
It'd be great to add a regression test for this as well.  There's a FakeDisposeSource in test/helper that is helpful in writing tests to verify disposal, and [here's an example of using it](https://github.com/cujojs/most/blob/7d27cd50afa2cacb6ee39fc62b424c73d61fe9e5/test/combinator/loop-test.js#L46) (there are other examples in the test folder as well).

briancavalier:
Looking good @ppoliani.  Thanks for adding the test.  I was porting your fix to `@most/core` in mostjs/core#107, and I noticed that there's actually another bug in the DebounceSink constructor.  It's not necessary to use `dispose.all` (or `disposeBoth` in `@most/core`), and the source disposable can just be assigned directly to `this.disposable`.

Take a look at mostjs/core#107 and see what you think.  If that seems right, then you can do the same here.

ppoliani:
@briancavalier The change makes sense 😄 

I'll include it here, as well.

briancavalier:
Awesome, thank you, @ppoliani.  We'll get this out in the next release soon.

PR #476, opened 2017-08-15
Clean up readme types

TrySound:
### Summary
Most itself support and fantasy and static land. Also monad already includes functor and applicative interfaces according both spec

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Most.js has a dependency on native Promises so a type definition for Promise mus
 <a href="https://github.com/fantasyland/fantasy-land"><img width="82" height="82" alt="Fantasy Land" src="https://raw.github.com/puffnfresh/fantasy-land/master/logo.png"></a>
 <a href="https://github.com/rpominov/static-land"><img width="131" height="82" src="https://raw.githubusercontent.com/rpominov/static-land/master/logo/logo.png" /></a>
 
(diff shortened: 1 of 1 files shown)


briancavalier:
Hey @TrySound, thanks!

My feeling is that even though it's redundant information, it's more helpful than harmful.  The current list is definitely inconsistent, though, since it enumerates Functor, Applicative, and Monad, but then only mentions Monoid rather than Semigroup and Monoid.

[Haskell docs](https://hackage.haskell.org/package/Stream-0.4.7.2/docs/Data-Stream.html#g:1) and [Purescript docs](https://pursuit.purescript.org/packages/purescript-lists/4.8.0/docs/Data.List#t:List) tend to include the exhaustive list (e.g. Semigroup and Monoid), although that's slightly different context since it's API documentation.

So, I guess we should do one of these to be consistent:

1. Merge this PR, to mention only the "top" of each hierarchy, or
2. Add Semigroup, to mention all of the type classes specifically

I'm leaning toward 2, since it feels like it does no harm, and may even be helpful to some.

Thoughts?

Frikki:
Too much information can be confusing, especially for newcomers. However, though FP is quite a difficult topic to approach, because of the mathematical names, it may be a good idea to be exhaustive and mention all the type classes because it provides the information that all type classes have been implemented and not just (wrongly) a subset, which sometimes is seen during the maturing of libraries. Thus, I also favor no. 2.

briancavalier:
> Too much information can be confusing, especially for newcomers.

Good point--always important to keep in mind!

PR #477, opened 2017-08-15
Throw error if handler is not present

TrySound:
### Summary

This example swallows errors
```js
const stream = most.throwError(1);
stream.subscribe({
  next: v => console.log(v)
});
```

### DIFF
--- a/src/observable/subscribe.js
+++ b/src/observable/subscribe.js
@@ -34,13 +34,23 @@ SubscribeObserver.prototype.event = function (t, x) {
 SubscribeObserver.prototype.end = function (t, x) {
   if (!this.disposable.disposed) {
     var s = this.subscriber
-    doDispose(this.fatalError, s, s.complete, s.error, this.disposable, x)
+    var fatalError = this.fatalError
+    Promise.resolve(this.disposable.dispose()).then(function () {
+      if (typeof s.complete === 'function') {
+        s.complete(x)
+      }
+    }).catch(function (e) {
+      throwError(e, s.error, fatalError)
+    })
   }
 }
 
 SubscribeObserver.prototype.error = function (t, e) {
   var s = this.subscriber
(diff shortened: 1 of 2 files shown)


TrySound:
Done

src/observable/subscribe.js
```diff
@@ -34,13 +34,23 @@ SubscribeObserver.prototype.event = function (t, x) {
...
   if (!this.disposable.disposed) {
     var s = this.subscriber
-    doDispose(this.fatalError, s, s.complete, s.error, this.disposable, x)
+    var fatalError = this.fatalError
+    Promise.resolve(this.disposable.dispose()).then(function () {
+      if (typeof s.complete === 'function') {
+        s.complete(x)
+      }
+    }).catch(function (e) {
+      throwError(e, s.error, fatalError)
```
briancavalier: Since a subscriber is an object, I think we should always call `s.error` with `s` as the `thisArg`.  I'm thinking we could pass `s` here instead of `s.error`.

Also, it looks like `throwError` is only call with `fatalError` as it's 3rd parameter (and it's hard to imagine calling it with anything else, in the current setup).  So, it may make sense just to remove the parameter and call `fatalError` directly in throwError.

Maybe something like:

```js
function throwError (e1, subscriber) {
  if (typeof subscriber.error !== 'function') {
    try {
      subscriber.error(e1)
    } catch (e2) {
      fatalError(e2)
    }
  } else {
    fatalError(e1)
  }
}

TrySound: Nope. It will make testing impossible.

briancavalier: Ah, yes, I see. Thanks for clarifying.

In that case, I think I'd still prefer passing the subscriber as the 2nd arg rather than using `bind` in 2 places.

PR #478, opened 2017-08-19
Add flow definitions

TrySound:


briancavalier:
Thanks @TrySound!  I started looking through this a bit tonight.  I should be able to get through all of it over the next few days.

type-definitions/most.js.flow
```diff
@@ -0,0 +1,410 @@
...
+  since(signal: Stream<any>): Stream<A>;
+  skipUntil(signal: Stream<any>): Stream<A>;
+  during(timeWindow: Stream<Stream<any>>): Stream<A>;
+  throttle(period: number): Stream<A>;
+  debounce(period: number): Stream<A>;
+
+  timestamp(): Stream<TimeValue<A>>;
+  delay(dt: number): Stream<A>;
+
+  await<B>(this: Stream<Promise<B>>): Stream<B>;
```
briancavalier: Afaik, using `this` here is valid in TS, but not in flow.  See https://github.com/facebook/flow/issues/452.

Unfortunately, I don't think flow has any way of handling this situation.  So, I think we either have to treat await() and awaitPromises() like join(), or _omit_ the method versions from the type definitions, forcing flow users to call the function versions instead.

Thoughts?

type-definitions/most.js.flow
```diff
@@ -0,0 +1,410 @@
...
+  multicast(): Stream<A>;
+
+  thru<B>(transform: (stream: Stream<A>) => B): B;
+}
+
+declare export function just<A>(a: A): Stream<A>;
+declare export function of<A>(a: A): Stream<A>;
+declare export function empty(): Stream<any>;
+declare export function never(): Stream<any>;
+declare export function from<A>(as: A[] | Iterable<A> | Observable<A>): Stream<A>;
```
briancavalier: At first, I thought `A[]` isn't valid flow array syntax, but [turns out it is](https://flow.org/en/docs/types/arrays/#toc-array-type-shorthand-syntax)!  TIL 😄 :+1:

type-definitions/most.js.flow
```diff
@@ -0,0 +1,410 @@
...
+  streams: [Stream<A>, Stream<B>, Stream<C>, Stream<D>]
+): Stream<R>;
+declare export function combineArray<A, B, C, D, E, R>(
+  fn: (a: A, b: B, c: C, d: D, e: E) => R,
+  streams: [Stream<A>, Stream<B>, Stream<C>, Stream<D>, Stream<E>]
+): Stream<R>;
+declare export function combineArray<V, R> (
+  fn: (...items: V[]) => R,
+  items: Stream<V>[]
+): Stream<R>;
```
briancavalier: Whatever we decide about any vs V for the method form of combineArray, we can apply here, too.

type-definitions/most.js.flow
```diff
@@ -0,0 +1,410 @@
...
+  tap(f: (a: A) => any): Stream<A>;
+  chain<B>(f: (a: A) => Stream<B>): Stream<B>;
+  flatMap<B>(f: (a: A) => Stream<B>): Stream<B>;
+
+  ap<B, C>(fs: Stream<(a: A) => B>): Stream<C>;
+
+  // Note: Without higher-kinded types, the types for these
+  // cannot be written in a completely safe manner; See https://github.com/Microsoft/TypeScript/issues/1290
+  // For better type safety, consider using most.join/switch/switchLatest with thru
+  join(): A;
```
TrySound: Oops. Skipped

type-definitions/most.js.flow
```diff
@@ -0,0 +1,410 @@
...
+    streams: [Stream<B>, Stream<C>, Stream<D>]
+  ): Stream<R>;
+  combineArray<B, C, D, E, R>(
+    fn: (a: A, b: B, c: C, d: D, e: E) => R,
+    streams: [Stream<B>, Stream<C>, Stream<D>, Stream<E>]
+  ): Stream<R>;
+  combineArray<V, R>(
+    fn: (a: A, ...rest: V[]) => R,
+    streams: Stream<V>[]
+  ): Stream<R>;
```
TrySound: Did you ever have cases with more then five streams?

TrySound: I don't think `any` is a good case. It will hide output type and skip errors when user trust flow. I saw typings with even `...rest: Array<void>`, so I vote for safer case.

package.json
```diff
@@ -15,7 +15,10 @@
   "scripts": {
     "unit-test": "buster-test",
     "test": "eslint src test && npm run unit-test && npm run doctest",
-    "build": "rimraf lib dist && buba src -o lib && rollup -c && uglifyjs dist/most.js -c \"warnings=false\" -m -o dist/most.min.js",
+    "build": "npm run build:dist && npm run build:min && npm run build:flow",
+    "build:dist": "rimraf lib dist && buba src -o lib && rollup -c",
+    "build:min": "uglifyjs dist/most.js -c \"warnings=false\" -m > dist/most.min.js",
+    "build:flow": "cpy type-definitions/most.js.flow dist",
```
TrySound: Moved to `src/index.js.flow`

type-definitions/most.js.flow
TrySound: Removed comment and methods

type-definitions/most.js.flow
TrySound: Removed methods

type-definitions/most.js.flow
TrySound: Good!

TrySound:
I think we can keep array definitions as is and maybe extend them when somebody will leave an issue.

type-definitions/most.js.flow
TylorS: I think `Stream<V>[]` is probably a bit restrictive. I agree with @briancavalier that `Array<Stream<any>>`.

If we're going to do any modifications to the TypeScript definitions adding a `ReadonlyArray<Stream<any>>` or a variant of that would be a nice addition

TrySound: I'm afraid this will break safety of the library even more, since we mentally trust to definitions.

TrySound: The problem is that typescript requires you to specify more types, but flow mostly can infer them and if it does not ask to specify type (it's not exported) then everything should work correctly.

briancavalier:
After thinking about this a bit more, I agree, @TrySound.  The current flow array defs match the TS defs, for better or for worse, at least they are consistent.  It'd also be a breaking change in 1.x to remove the catch-all TS variants, so I think we just go with it in 1.x and think ahead to a better solution for 2.x.

src/index.js.flow
```diff
@@ -0,0 +1,403 @@
+// @flow
+
+declare type SeedValue<S, V> = { seed: S, value: V };
+declare type TimeValue<V> = { time: number, value: V };
+
+declare type CreateGenerator<A> = (...args: Array<any>) => Generator<A|Promise<A>, any, any>;
+
+export interface Sink<A> {
```
briancavalier: One last thing: I'd like to go with types unless there's a compelling reason to prefer interfaces.

We've [preferred `type` over `interface`](https://github.com/mostjs/core/blob/master/packages/types/index.js.flow) in `@most/core`'s flow types.  I don't see any benefit to having 2 things, types and interfaces, vs having only types.  It'd also nice to be consistent with `@most/core`, although I'm not too concerned about it.

Otherwise, this all looks good to me.

PR #479, opened 2017-08-22
Mention zero and foldable algebraic types

TrySound:
[Plus](https://github.com/rpominov/static-land/blob/master/docs/spec.md#plus)::zero is implemented via `empty`, `never` methods
Not sure about `scan`, but `reduce` looks like [Foldable](https://github.com/rpominov/static-land/blob/master/docs/spec.md#foldable)::reduce

We can add some examples with laws around these types.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -17,10 +17,10 @@ export { Stream }
 
 // Add of and empty to constructor for fantasy-land compat
 Stream.of = of
-Stream.empty = empty
+Stream.empty = Stream.zero = empty
 // Add from to constructor for ES Observable compat
 Stream.from = from
-export { of, of as just, empty, never, from, periodic }
+export { of, of as just, empty, empty as zero, never, from, periodic }
 
 // -----------------------------------------------------------------------
 // Draft ES Observable proposal interop
@@ -265,7 +265,7 @@ Stream.prototype.join = function () {
 import { continueWith } from './combinator/continueWith'
 
 // @deprecated flatMapEnd, use continueWith instead
-export { continueWith, continueWith as flatMapEnd }
+export { continueWith, continueWith as alt, continueWith as flatMapEnd }
 
 /**
  * Map the end event to a new stream, and begin emitting its values.
@@ -278,6 +278,7 @@ Stream.prototype.continueWith = function (f) {
   return continueWith(f, this)
 }
 
+Stream.prototype.alt = Stream.prototype.continueWith
 // @deprecated use continueWith instead
 Stream.prototype.flatMapEnd = Stream.prototype.continueWith
 
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -77,6 +77,7 @@ declare export class Stream<A> {
   ap<B, C>(fs: Stream<(a: A) => B>): Stream<C>;
 
   continueWith(f: (a: any) => Stream<A>): Stream<A>;
+  alt(f: () => Stream<A>): Stream<A>;
   concatMap<B>(f: (a: A) => Stream<B>): Stream<B>;
   mergeConcurrently<B>(concurrency: number): Stream<B>;
   merge(...ss: Array<Stream<A>>): Stream<A>;
@@ -210,6 +211,7 @@ declare export class Stream<A> {
 declare export function just<A>(a: A): Stream<A>;
 declare export function of<A>(a: A): Stream<A>;
 declare export function empty(): Stream<any>;
+declare export function zero(): Stream<any>;
 declare export function never(): Stream<any>;
 declare export function from<A>(as: A[] | Iterable<A> | Observable<A>): Stream<A>;
 declare export function periodic<A>(period: number, a?: A): Stream<A>;
@@ -237,6 +239,7 @@ declare export function join<A>(s: Stream<Stream<A>>): Stream<A>;
 declare export function switchLatest<A>(s: Stream<Stream<A>>): Stream<A>;
 
 declare export function continueWith<A>(f: (a: any) => Stream<A>, s: Stream<A>): Stream<A>;
+declare export function alt<A>(f: () => Stream<A>, s: Stream<A>): Stream<A>;
 declare export function concatMap<A, B>(f: (a: A) => Stream<B>, s: Stream<A>): Stream<B>;
 declare export function mergeConcurrently<A>(concurrency: number, s: Stream<Stream<A>>): Stream<A>;
 
--- a/test/alt-plus-test.js
+++ b/test/alt-plus-test.js
@@ -1,0 +1,58 @@
+/* global describe, it */
+require('buster').spec.expose()
+const { expect } = require('buster')
+
+const most = require('../src')
+
+const reduceOne = s => s.reduce((acc, x) => [...acc, x], [])
+
+const reduceTwo = (s1, s2) => Promise.all([reduceOne(s1), reduceOne(s2)])
+
+describe('Alt', () => {
+  it('should satisfy associativity', () => {
+    const s1 = most.of(1).continueWith(() => most.of(2)).continueWith(() => most.of(3))
+    const s2 = most.of(1).continueWith(() => most.of(2).continueWith(() => most.of(3)))
+
+    return reduceTwo(s1, s2).then(([r1, r2]) => {
+      expect(r1).toEqual(r2)
+    })
+  })
+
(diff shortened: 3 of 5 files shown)


briancavalier:
Hi @TrySound.  As it stands, I don't think we can mention either Plus or Foldable.  I'll try to explain why for each.

#### Foldable

If you look carefully at the type definition of Foldable:

```hs
reduce :: Foldable f => f a ~> ((b, a) -> b, b) -> b
```

you'll notice that given a function `(b, a) -> b`, `reduce` must be able to return a `b` _synchronously_.  That's impossible in the general case with an asynchronous data structure like a mostjs event stream.  The best it can do in the general case is to return a future `b`, iow a promise for `b`, which violates Foldable.

#### Plus

Plus [requires](https://github.com/fantasyland/fantasy-land#plus) the type first to implement [Alt](https://github.com/fantasyland/fantasy-land#alt):

> A value that implements the Plus specification must also implement the Alt specification.

So, since mostjs doesn't currently implement Alt, it wouldn't be valid, for example, to alias empty() or never() to zero() and claim support for Plus.

That raises the question: could we implement Alt?  The core team has discussed this, and if memory serves, we believe that there are at least two possible Alt implementations, each of which implies a different Plus implementation:

1. Alt is a [race](https://github.com/cujojs/most/wiki/Race).  This implies never() as the correct Plus zero().
2. Alt is a continuation, i.e. "stream 1 until it ends, and then stream 2", which is sort of like continueWith or concat.  This implies empty() as the correct Plus zero().  There is a precedent for this in that [netwire]'s Alternative instance is similar.

It's not clear which of these is the more useful and/or correct Alt/Plus.  My most recent thinking has been that, since `concat` was removed in `@most/core`, it'd be reasonable to introduce Alt as option 2 above, with empty() as Plus zero().  But, honestly, it's still not entirely clear what is best.

Any thoughts on that?  Do you see other options for a valid and useful Alt?

TrySound:
Was concat remove because of event time breaking?
I like the idea of empty/continueWith pair

briancavalier:
> Was concat remove because of event time breaking?

Yeah, and it just has a misleading mental model.  You can read more here: https://github.com/mostjs/core/pull/80

TrySound:
So do we leave plus and alt here?

briancavalier:
Yeah, I'm up for adding them.  I'd like to hear what @TylorS, @davidchase, and @Frikki think about it.

If we add them, this PR will need a few more todos:

- [ ] Implementations.  These are trivial, just aliasing, really, e.g. `Stream.zero = empty`,
- [ ] Tests.  Probably simple tests for the aliases [here](https://github.com/cujojs/most/blob/master/test/most-test.js) plus tests specifically for the [Alt](https://github.com/fantasyland/fantasy-land#alt) and [Plus](https://github.com/fantasyland/fantasy-land#plus) laws.  Two of the Plus laws are already tested [here](https://github.com/cujojs/most/blob/master/test/build-test.js#L44-L64)
- [ ] API Documentation. It'd be great to describe them in a way that distinguishes them, conceptually, from concat, since concatenation is a problematic concept.

If we all agree this is the right direction, then I'll be happy to help.

briancavalier:
After some offline discussion, I think we're good to go on this.  @TrySound, are you up for handling the todos above?

TrySound:
Sure.

briancavalier:
Awesome, thank you!

TrySound:
@briancavalier What about monoid/semigroup types? They look similar, but removing concat can break semigroup.

briancavalier:
Semigroup and Monoid are similar to Alt and Plus.  Semigroup is a superclass of Monoid (see [here](https://github.com/fantasyland/fantasy-land#monoid) and [diagram here](https://github.com/fantasyland/fantasy-land/blob/master/figures/dependencies.png)).  IOW, a type must first implement Semigroup before it can be a Monoid.  So, if concat is removed, then it's no longer a Semigroup, and thus, by definition, no longer a Monoid.

The two hierarchies are similar, but conceptually, it can be helpful sometimes to think of Semigroup/Monoid as representing merging or combining, and Alt and Plus as representing choice or alternatives.

TrySound:
You use `merge` word here. So can be semigroup used, not like `appending` in the end, but combining values in some way?

briancavalier:
@TrySound English is imprecise, so in reality, the only things that matter are the type signature and mathematical laws by which concat must abide:

```
concat :: forall s. (Semigroup s) => s -> s -> s
```

You can substitute `Stream a` for `s` and get:

```
concat :: forall a. Stream a -> Stream a -> Stream a
```

That is, given 2 streams of `a`s, produce another stream of `a`s.  And there is an associativity law:

```
concat(s1, concat(s2, s3)) ~= concat(concat(s1, s2), s3)
```

Appending and time-based interleaving are both valid implementations.  It is possible to have a Semigroup that combines values, but only in the case where you know the values (i.e. the `a`s are also a Semigroup).  For example, it's typical for a Maybe structure, such as [purescript's Maybe](https://github.com/purescript/purescript-maybe/blob/master/src/Data/Maybe.purs#L176), where `Maybe a` is a Semigroup only if `a` is a Semigroup.

An analogy for an event stream would be a merge operation that requires `a` to be a Semigroup and `concat`'s the `a`s when events are simultaneous in the two streams being merged.

Currently, most.js's `concat` appends one stream to another.  This was a bad decision for a data structure in which times are just as important as values, because appending causes time shifting.  It's likely in most.js 2.0, we'll switch the fantasy-land Semigroup to use merge() instead, which preserves times.

TrySound:
Added separate test file for Alt and Plus (do you want to split them?).
Covered `continueWith`, `alt`, `empty`, `zero` with type tests. Do we need any parameter in `continueWith`? Is it kind of legacy?
Is it ok to use es features in tests?

TrySound:
Nothing is deprecated in this PR. Just aliases for algebraic types like `just/of`.

Frikki:
My bad, @TrySound. The deprecate comments were already there. However, I’ll then alert @briancavalier on this practice.

src/index.js
```diff
@@ -265,7 +265,7 @@ Stream.prototype.join = function () {
 import { continueWith } from './combinator/continueWith'
 
 // @deprecated flatMapEnd, use continueWith instead
-export { continueWith, continueWith as flatMapEnd }
+export { continueWith, continueWith as alt, continueWith as flatMapEnd }
```
briancavalier: We can't use `alt` as an alias for `continueWith`, since they have different types: `alt` accepts a Stream, and `continueWith` accepts a function that returns a Stream:

```hs
alt :: Stream a -> Stream a -> Stream a
continueWith :: Stream a -> (() -> Stream a) -> Stream a
```

However, `alt` _can_ be an alias for [`concat`](https://github.com/cujojs/most/blob/master/src/combinator/build.js#L23).  Note that `concat` is deprecated, which is ok.  We can remove the name `concat` in 2.0 and still keep `alt`.

TrySound: But `alt` then will be incorrect for streams like `concat` right now.

TrySound: Why should we keep it?

briancavalier: Sorry, it was probably confusing that I said `alt` should be an alias for `concat`.  That implies that they have the same intention.  In my view, they don't have the same intention, even though their _implementations_ can be identical _temporarily_.

#### The problem with concat

The problem with concat, [as we've discussed](https://github.com/mostjs/core/pull/80), is the mental model of "concatenation", which tends to be associated with arrays and other index-based data structures.  That intuition doesn't work well for time-sensitive structures, where concatenation implies time-shifting.

#### The intention of alt

So, even if alt has the same _implemention_ as concat in 1.0, it has a different intention, that of "alternative", and we can (and should!) document it differently.  Rather than concatenation, the intention of alt is to observe one stream until it ends and then begin observing the other alternative stream.  Haskell's Netwire is my current mental model/precedent for alt.

#### Should we keep alt in most 1.x?

I don't mind including it in 1.x with carefully worded documentation.  If folks feel like it is confusing and not useful, the we probably shouldn't have alt in 1.x.

In most 2.0, which will be based on `@most/core`, I think alt will be more intuitive.  

#### Should we delete concat in 2.0?

From the perspective of a Semigroup, if we continue Fantasyland support in 2.0, IMHO we should change `concat` to be an alias for a binary `merge` operation (and keep `alt` as above)

PR #484, opened 2017-08-29
Allow omitting properties of subscriber

TrySound:
### Summary

According [spec](https://github.com/tc39/proposal-observable#observer) all props of subscriber can be omitted. Also I added first test case for subscriber and would like to improve coverage step by step.

### DIFF
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -48,10 +48,10 @@ export type Observable<A> = {
 }
 
 export type Subscriber<A> = {
-  next(value: A): void;
-  error(err: Error): void;
+  next?: (value: A) => void;
+  error?: (err: Error) => void;
   // complete value parameter is deprecated
-  complete(value?: A): void;
+  complete?: (value?: A) => void;
 }
 
 export type Subscription<A> = {
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -48,10 +48,10 @@ export type Observable<A> = {
 }
 
 export type Subscriber<A> = {
-  next?: (value: A) => void;
-  error?: (err: Error) => void;
+  +next?: (value: A) => void;
+  +error?: (err: Error) => void;
   // complete value parameter is deprecated
-  complete?: (value?: A) => void;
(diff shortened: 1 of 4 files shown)


src/index.js.flow
```diff
@@ -48,10 +48,10 @@ export type Observable<A> = {
 }
 
 export type Subscriber<A> = {
-  next(value: A): void;
-  error(err: Error): void;
+  next?: (value: A) => void;
+  error?: (err: Error) => void;
   // complete value parameter is deprecated
-  complete(value?: A): void;
+  complete?: (value?: A) => void;
```
briancavalier: I've had problems in the past where the property style (vs. method style) object type definition prevents class instances from satisfying the type.  The reason is that flow requires the properties to be own props, but class methods are inherited (on the prototype).

There's some related info [here](https://github.com/facebook/flow/issues/2770), [here](https://github.com/facebook/flow/issues/3076), and [here](https://flow.org/blog/2016/10/04/Property-Variance/)

One way to verify is to create a class that has a `next` method and see if it type checks as a `Subscriber`.  Something like:

```js
class Test {
  next (a: number) void {}
}

const s: Subscriber<number> = new Test()
```

If that doesn't work, I think one solution may be to make the properties covariant explicitly:

```js
export type Subscriber<A> = {
  +next?: (value: A) => void;
  +error?: (err: Error) => void;
  // complete value parameter is deprecated
  +complete?: (value?: A) => void;
}
```

TrySound:
Done. Covered with test.

type-definitions/most-flow-test.js
```diff
@@ -0,0 +1,22 @@
...
+
+subscribe({}, just(1))
+subscribe({ next: d => { (d: number) } }, just(1))
+subscribe({ error: err => { (err: Error) } }, just(1))
+subscribe({ complete: () => {} }, just(1))
+const { unsubscribe } = subscribe(new class {
+  next() {};
+  error() {};
+  complete() {};
+}, just(1))
```
briancavalier: Nice

PR #494, opened 2017-09-22
Remove await and awaitPromises methods

TrySound:
These methods depends on `this` indentifier which is forbidden in flow 0.55. Same as `join`


TrySound:
Hi, @briancavalier . Can you take a look? This makes definitions broken with flow55

PR #506, opened 2017-11-24
Fix typings for unfold

eddiew:
### Summary
The typings for Most.unfold don't match what the code does or the API docs, so I fixed them.

### DIFF
--- a/type-definitions/most.d.ts
+++ b/type-definitions/most.d.ts
@@ -1,4 +1,5 @@
 declare type SeedValue<S, V> = { seed: S, value: V };
+declare type UnfoldValue<S, V> = SeedValue<S,V> | { done: boolean }
 declare type TimeValue<V>    = { time: number, value: V };
 
 declare interface Generator<A, B, C> {}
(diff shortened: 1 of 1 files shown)


briancavalier:
Hey @eddiew, thanks.  Hmmm, yes, the type seems wrong.  I like your idea to introduce an `UnfoldValue` sum type.  How about this?

```js
type UnfoldValue<S, V> = { done: false, seed: S, value: V } | { done: true, value?: V }
```

The first variant could use an intersection with the existing SeedValue type--that'd be fine.  It needs to have a `done: boolean`, so maybe like:

```js
type UnfoldValue<S, V> = { done: false } & SeedValue<S, V> | { done: true, value?: V }
```

I think using an intersection is a bit harder to read, though.  What do you think?

eddiew:
I thought `done` was allowed to be undefined if not `true`, so then it wouldn't be necessary to have the `done: false` in the first variant. In that case we could do:

`type UnfoldValue<S, V> = SeedValue<S, V> | { done: true, value?: V }`

But if you'd rather keep the `done: false` then I agree that the intersection is a little bit confusing, and I like the first version you have better.

MaxSvargal:
Maybe also add transduce to definitions?

briancavalier:
Sorry for taking a while to respond!

> I thought done was allowed to be undefined if not true

Hmmm, yes, the current docs do indeed read that way, and so does the example.  Ok, let's go with allowing done to be missing or undefined.  Thanks for pointing that out.

briancavalier:
@MaxSvargal Let's do that in another PR.  See also: https://github.com/cujojs/most/issues/512 where it's being discussed.

type-definitions/most.d.ts
```diff
@@ -1,4 +1,5 @@
 declare type SeedValue<S, V> = { seed: S, value: V };
+declare type UnfoldValue<S, V> = SeedValue<S,V> | { done: boolean }
```
briancavalier: Since the docs are kind of loose, we may need to be even more permissive.  Maybe we can go with this, and then relax it if it causes someone some pain.

Seems like we should tighten the done case to `{ done: true }` since it'd be pointless to return `{ done: false }`.  What do you think?

eddiew:
Good catch. I updated the code to require `done: true`

PR #510, opened 2017-12-01
Use standard, update deps, drop some

briancavalier:
### Summary

Update deps, simplify build, switch to standard, fix lint.  Drop node 4, add node 8 to travis.

Lots of deps had gotten out of date and crusty.  Also, buster requires contextify, which doesn't build on Node 4.  This attempts to clean up as much of that situation as possible, so that travis is happy.


briancavalier:
This PR doesn't "fix" the contextify warnings.  Buster still requires it, and it building it generates warnings on node 6 and 8.  It _does_ build successfully, but looks horrible in the console.

PR #517, opened 2018-02-09
removed header-breaking anchor tag

jaawerth:
### Summary

Removed a dangling anchor tag that was breaking the header for ES Observable interop

### Todo

Should be good to go!


briancavalier:
Thanks!

PR #520, opened 2018-06-07
fix(types): Added flow type definitions for await and awaitPromises.

critocrito:
### Summary

Using `most` with function chaining doesn't type check using [`flow`](https://flow.org) when using `awaitPromises` with the following error: `Cannot call fromEvent(...).map(...).awaitPromises because property awaitPromises is missing in Stream`. The fix adds a type definition for the `await` and `awaitPromises` property and allows flow to type check.

I saw that the original type definitions were removed in #494. From my understanding flow doesn't require a type definition for `this`. But since I'm fairly new to flow I might be missing something. I use flow version 0.73.

### DIFF
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -154,6 +154,9 @@ declare export class Stream<A> {
   timestamp(): Stream<TimeValue<A>>;
   delay(dt: number): Stream<A>;
 
+  await<B>(): Stream<B>;
+  awaitPromises<B>(): Stream<B>;
+
   sample<B, C, R>(
     fn: (b: B, c: C) => R,
     b: Stream<B>,


briancavalier:
Hi @critocrito.  The problem is that the awaitPromises _method_ should only be callable when `this` is `Stream<Promise<A>>`.  Unfortunately, while Flow has good _inference_ of `this`, it lacks the capability to specify _constraints_ on `this`.  That means Flow would silently hide a mistake like:

```js
const s1 = most.of(Promise.resolve(1))
const s2: Stream<string> = s1.awaitPromises() // Ideally would fail, but doesn't
```

Since one goal of using a type system should be to prevent such mistakes, we removed the method type definition.  It will be great if Flow adds support for `this` constraints.  Please voice your support here: https://github.com/facebook/flow/issues/452

As a workaround, you can use the [function version of awaitPromises](https://github.com/cujojs/most/blob/master/src/index.js.flow#L321), which does provide the proper constraint.  You can still have a fluent API by using [thru](https://github.com/cujojs/most/blob/master/docs/api.md#thru):

```js
fromEvent(...).map(...).thru(awaitPromises)
```

Hope that helps!

briancavalier:
Happy to help, @critocrito.

PR #523, opened 2018-08-09
test performance upgrade

cortopy:
### Summary

New performance tests are in!!
Major changes:
- Added rxjs6
- Upgraded all perf test dependencies
- Run tests in node 8

### Todo
Compare most.js with the latest versions of other reactive frameworks

### DIFF
--- a/test/perf/merge-nested.js
+++ b/test/perf/merge-nested.js
@@ -2,8 +2,9 @@ require('buba/register')
 var Benchmark = require('benchmark');
 var most = require('../../src/index');
 var rx = require('rx');
-var rxjs = require('@reactivex/rxjs')
-var kefir = require('kefir');
+var rxjs = require('@reactivex/rxjs');
+var rxjs6 = require('rxjs');
+var rxjs6Operators = require('rxjs/operators');
 var bacon = require('baconjs');
 var highland = require('highland');
 var xs = require('xstream').default;
@@ -33,6 +34,14 @@ function merge(n, create) {
     s = s.merge(create(a))
   }
   return s;
+}
+
+function mergeRxjs6(n, create) {
+  var s = [];
+  for(var i=0; i<n; ++i) {
+    s.push(create(a));
+  }
+  return create(a).pipe(rxjs6Operators.merge(...s));
 }
 
 function mergeHighland(n) {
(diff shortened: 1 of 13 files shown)


briancavalier:
Thanks, @cortopy! I should have time to look at this over the weekend.

briancavalier:
Quick, question about how you ran these.

It's best to run them with as little else running on the machine as possible.  There's really no way to get a "perfect" run (whatever that might mean), of course, but at least on OS X I've seen some weird outliers when I've forgotten to quit everything but a terminal window.

So, if you did that, great!  This looks ready to merge.  If not, would you mind doing another run?  Thanks!

Also, if you're up for it, I'd be delighted to have a similar PR over at [`mostjs/core`](https://github.com/mostjs/core/tree/master/packages/core/test/perf).

cortopy:
I did close all windows for the reasons mentioned. But that's a good point. I'll try again by restarting my computer in terminal mode

cortopy:
@briancavalier interestingly, not much of a difference in terminal mode. I've done a new commit with the results

briancavalier:
Awesome, thanks @cortopy.  Now we know :). I guess operating systems (perhaps particularly linux) have gotten pretty good at not using resources when idle.

test/perf/merge-nested.js
```diff
@@ -35,6 +36,14 @@ function merge(n, create) {
   return s;
 }
 
+function mergeRxjs6(n, create) {
+  var s = [];
+  for(var i=0; i<n; ++i) {
+    s.push(create(a));
+  }
+  return create(a).pipe(rxjs6Operators.merge(...s));
```
briancavalier: I don't think this is testing the same intent, and I apologize if that intent wasn't clear!  The intent is specifically to merge pairs of streams, rather than to rely on a library's bulk merge operator.

The reason is that across a large application, you may not tend to have all the streams in your hand at once.  For example, you might call a function which, unbeknownst to you, internally merges 2 streams, and returns the result to you, and you then merge that with another stream, and so on.

MostJS focuses not only on performance, but performance scalability, ie how does it perform as usage scales _without_ the dev having to think about it.  For example, MostJS can (and does) optimize the situation described above.

So, I think rx 6 can reuse the `merge` helper function above, and provide a small wrapper lambda around `rxjs6.from` as `create`, similar to what the rx 5 tests do.

Does that make sense?

PR #545, opened 2020-10-07
Fix build by using require for new cjs deps

briancavalier:
### Summary

Fix build by using require for new cjs deps after https://github.com/cujojs/most/pull/541

It's a bummer, but our antiquated build setup is failing because it can't handle some kinds of CommonJS / ES module interop.  This is the simplest short term fix I could get working.  Here's what `npm run build` says:

Screen Shot 2020-10-07 at 6 27 47 PM

After this change:

Screen Shot 2020-10-07 at 6 30 25 PM

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,8 +9,8 @@ import * as base from '@most/prelude'
 import { of, empty, never } from './source/core'
 import { from } from './source/from'
 import { periodic } from './source/periodic'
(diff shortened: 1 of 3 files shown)


Frikki (approved):
Well, isn’t that something?

TylorS (approved):

